Issue #40, opened 2015-11-08
Separate the HAL from the mini-OS/Touch-develop runtime


markshannon:
The DAL currently consists of a genuine hardware abstraction layer (HAL) and an assortment of lightweight OS and language runtime features.
Unfortunately it is currently impossible to use the  HAL without getting all the other stuff.
For MicroPython and, I suspect other systems, it would be highly desirable to have access to the HAL without having to work around the other features.

For the specific case of MicroPython on the micro:bit we need the following features that the DAL provides:
* reset
* I2C
* PWM
* Pins
* Pin de-bouncing (currently residing in the Button class)
* Compass
* Accelerometer
* Font
* UART/serial (although the underlying mbed class may be sufficient)

We *don't* need the following features:
* Heap allocator
* Fibers
* Events, Message Bus and Listeners
* Managed Strings
* Image
* Button (apart from the de-bouncing logic)
* MultiButton

I've deliberately not mentioned the display or BLE on either list.
MicroPython already has its own implementation of the display, so we don't need the DAL display. 
However it might save some code duplication if the lowest level functionality of the display were exposed. A stand alone `panic()` function would also be useful.

Separating out the BLE related parts will be more complex, and best left for later.

Ideally we will be able to replace `#include "MicroBit.h"` with `#include "MicroBitHal.h` in the MicroPython code and save ourselves the overhead of the unwanted features.

jamesadevine:
Hi @markshannon, we have now merged in the component-refactor branch which leads to a componentised DAL. Happy to receive comments regarding any further changes you think are required!

Closing this one for now though!

PR #42, opened 2015-11-08
Move PWM code into HAL and factor out debouncing from MicroBitButton

markshannon:
A first step to fixing https://github.com/lancaster-university/microbit-dal/issues/40
Move PWM into `hal/` sub folder.
Separate out the debouncing aspect of `MicroBitButton` from the event signalling and move it into the HAL.


jamesadevine:
Superseded by recent changes to the DAL.

Closing.

markshannon:
It hasn't been superseded. 
The MicroBitButton is still coupled to the MicroBitSystemTimer, so we can't use it independently.

finneyj:
Hello @markshannon.

Why is that a problem for you? MicrobitSystemTimer is just a lightweight multiplexed Ticker, so we can save some RAM (lots of components need a sense of time, and Ticker instances are quite expensive in terms RAM).

I thought it was the fiber scheduler and message bus that was causing the problems?

markshannon:
But why depend on it at all?

finneyj:
oh, that's for the debounce logic and event timers (to detect a button HOLD for example). You can't debounce without some notion of the passage of time.

markshannon:
The `DebouncedPin` class in this PR has a `tick` method, but does not enforce the use of any particular timer.
So, why does depend `MicrobitButton.cpp` depend on "MicroBitSystemTimer.h"?

finneyj:
Iin my mind this is a case of simple encapsulation - the timer is an internal requirement for the component so it takes care of it transparently. As a periodic timer is a common requirement for quite a few components, having a multiplexed clock source makes sense to reduce RAM overhead. I'm still not clear what the problem is here.

It feels like we're getting a bit caught up on a specific detail here though. Both microbit-dal and micropython have moved on considerably since we last talked on this thread. Would be not be better off discussing things afresh from a top down perspective, then go from there?

i.e. now we have a version of microbit-dal with a set of components that run without the scheduler, memory manager and messagebus, ble, etc, perhaps we should first be discussing the architecture for this to be best utilized by micropython?

finneyj:
what thinks @markshannon? Want to join forces on this?

Broadening our discussion could give us better solutions - e.g. if you have issues with timers, then a bit of work aligning on MicroBitSystemTimer would provide support for a whole set of components transparently?

markshannon:
@finneyj Sounds like a good idea, but I'm going to be *really* busy until early June. So progress will be slow to non-existent on my side until then.

finneyj:
cool. No worries @markshannon - totally understand. We're all working this around our other responsibilities... Just give me a shout when you have time to do a bit of design.

ntoll:
Perhaps it would be useful to get "brains in the same room" sometime in June/July? I always find face-to-face meetings result in good things.

finneyj:
Absolutely - co-design always best done around a whiteboard, with as much coffee as can be feasibly consumed!

Not sure where you guys are based, but we'd happily host if you'd like to visit.

ntoll:
I'm in Northampton (actually Towcester, but close enough), @markshannon is relatively close by in Oxford (i.e. I could give him a lift) and @dpgeorge is in Cambridge. Perhaps we could wangle something half way (approximately just north of Brum) so everyone could do it in a day over a weekend..?

markshannon:
In June/July I'll be in San Francisco, so it might be awkward to pop over to Cambridge :smile:

ntoll:
Ahh... well, at the start of June I'll be in Portland for PyCon. ;-) See you there! @finneyj why not come to Portland? ;-)

mmoskal:
Very good coffee and fine craft beer. And the only chance to see the place where young people go to retire before it is completely overrun by climate change refugees from California! ;)

PR #98, opened 2016-02-29
README.md: Update for modern yotta, gcc first

jaustin:
Quite a lot has chanced since the original instructions were written,
and so this commit updates the instructions to be a little more useful.

Most significantly, we now walk through the process with gcc not
armcc, as this is now the default toolchain for builds of the
micro:bit DAL by both Microsoft and Lancaster

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,59 +1,75 @@
 # microbit-dal
 
-## Building a project for the micro:bit using Yotta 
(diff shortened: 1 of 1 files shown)


jamesadevine:
@jaustin in microbit-docs.

Closing.

Issue #114, opened 2016-04-07
Notifications from Magnetometer Bearing characteristic do not seem to work


bluetooth-mdw:
Tested 1.4.17 with nRF Master Control Panel. Notifications are never received.

jamesadevine:
@martinwork could you retest against master?

martinwork:
Related to #100 and #78?

I think the problem was that the build I was testing (1.4.x) did not trigger a call to calibrate the compass. With firmware based on the current 2.0.0 master, calibration is triggered when the MicroBitMagnetometerService object is created during the first reset after flashing. After that the bearings arrive as expected. However the bearings seem quite random. I suspect I am not playing the game right!

jamesadevine:
@martinwork  - or you're in a noisy environment :smile:

Shall we close this then?

martinwork:
It was opened by the other Martin

jamesadevine:
Yes - we should wait for final confirmation from @bluetooth-mdw 

Thanks for testing @martinwork !

Issue #29, opened 2015-10-20
Explicitly terminate active BLE link when reset is pressed


ghost:
As discussed and relating to https://github.com/bbc/microbit-native-android/issues/39,  if technically feasible it would be good to send a link layer LL_TERMINATE_IND PDU from micro:bit to a connected peer when reset is pressed. This will explicitly tell the peer (e.g. smartphone) that the link is going down. At present I think the smartphone assumes the link is still active until a supervision time out occurs later on.

PR #120, opened 2016-04-15
Added BLE connect/disconnect events and fixed bug in MagnetometerServ…

bluetooth-mdw:
…ice re: bearing characteristic

See https://drive.google.com/file/d/0B9Z54U0dfwbFN0Y0RVVROS1DQWM/view?usp=sharing for testing of BLE connection events.


inc/bluetooth/ExternalEvents.h
```diff
@@ -33,6 +33,9 @@ DEALINGS IN THE SOFTWARE.
 #define MICROBIT_ID_BLE             1000
 #define MICROBIT_ID_BLE_UART        1001
 
+#define MICROBIT_BLE_CONNECTED      1
```
jamesadevine: Hi @bluetooth-mdw,

Great PR, just a few minor things

We have an unofficial connection in terms of event codes, for example:
`MICROBIT_COMPASS_EVT_CAL_END`

If both `MICROBIT_BLE_CONNECTED` and `MICROBIT_BLE_DISCONNECTED ` could be renamed that would be great!

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -267,6 +277,9 @@ void MicroBitBLEManager::init(ManagedString deviceName, ManagedString serialNumb
     // automatically restart advertising after a device disconnects.
     ble->gap().onDisconnection(bleDisconnectionCallback);
     ble->gattServer().onSysAttrMissing(bleSysAttrMissingCallback);
+    
+    // generate an event when a Bluetooth connection is lost
```
jamesadevine: I think this comment should read:

"generate an event when a Bluetooth device connects" or something along those lines.

remay: // generate an event when a Bluetooth connection is **established**

jamesadevine: @remay 's version is a lot better... :smile:

jamesadevine:
@bluetooth-mdw Am I right in thinking this will fix #114 as well?

bluetooth-mdw:
Hopefully https://github.com/lancaster-university/microbit-dal/issues/114 will be fixed too. I've not witnessed this issue though so maybe the issue reporter can re-test?

Only just realised there's a comment here about a code comment! Let me get on that too.

remay:
@jamesadevine I was going to point out that we change the BLE disconect handler when we reset() the device (see #29).  But I either can't find my way around the new code, or we've lost this functionality.  Can you point me in the right direction?

jamesadevine:
@remay great catch... https://github.com/lancaster-university/microbit/blob/master/inc/MicroBit.h#L418

Martin - could you also make a PR here as well?

jamesadevine:
@bluetooth-mdw - this looks good to me!

bluetooth-mdw:
My first contribution to the micro:bit runtime. The world will never be the same again! :-)

remay:
@jamesadevine @bluetooth-mdw 

> great catch... https://github.com/lancaster-university/microbit/blob/master/inc/MicroBit.h#L418

This implementation doesn't change the disconnect callback handler, so nothing to do.

jamesadevine:
@remay very true - thanks @bluetooth-mdw

PR #139, opened 2016-04-29
Microbit filesystem.

jamesadevine:
Rebased version of the filesystem, with a `MicroBitFile` wrapper, and minor modifications to `MicroBitFileSystem`.

### DIFF
--- a/inc/drivers/MicroBitFile.h
+++ b/inc/drivers/MicroBitFile.h
@@ -1,0 +1,533 @@
+#ifndef MICROBIT_FILE_H_
+#define MICROBIT_FILE_H_
+#include "MicroBitFlash.h"
+#include <stdint.h>
+
+/**
+  * @file MicroBitFile.h
+  * @author Alex King
+  * @date 6 Mar 2016
+  *
+  * @brief API for POSIX-like file system interface: open/close/read/write.
+  *
+  * -- Basic overview of how the file system works:
+  * All file meta data is stored in the first page. Each file has a record
+  * entry (struct FileTableEntry_t), which stores:
+  *  - Filename
+  *  - File size
+  *  - An ordered list of pages/blocks that store the file data itself.
+  * 
+  * The table housing these records is referred to as the File Table (FT)
+  * and looks something like this:
+  *  {"file1.txt",   5000,   {1,2,3,6}}  // 5000-byte file, in blocks 1,2,3,6
+  *  {"file02.txt",   200,   {5}}        // 200 byte file in a single block
+  * 
+  * The first FT entry is reserved, and is used only to store a list of 
+  * currently available blocks for writing.
+  * Thus to expand a file size, a block must be removed from this list, and 
+  * added to that files blocklist.
+  * 
+  * Since there can only be one FT, the number of entries is restricted
+  * to how many can fit in a single flash page. Hence, this is the 
+  * maximum number of files that can be stored.
+  * 
+  * The filesize is stored in the FT, but cached in the mb_fd struct.
+  * Writes to the FT filesize are made sparingly in write() and close(), to
+  * reduce erasures.
+  * @warning close() *must* be called to ensure the FT is synched, before 
+  * the microbit is finished.
+  * 
+  * -- API overview.
+  * Source code is divided logically into two parts:
+  * ft_*() functions - to access and modify the FT entries.
+  *                    E.g., to create a new file, change a files size,
+  *                    or append new blocks to the file.
+  *                    These are private methods internal to the class.
+  * 
+  * read/write/seek/remove - POSIX-style file access functions.
+  *                          These are the methods used to interact with the
+  *                          file system.
+  * 
+  * Example:
+  * @code
+  * MicroBitFile f;
+  * int fd = f.open("myFile.txt", MB_WRITE|MB_CREAT);
+  * if(fd < 0) {
+  *   print("couldn't open a new file");
+  *   exit;
+  * }
+  * if(f.write(fd, "Hello, World!\n",14) != 14) {
+  *   print("error writing!")
+  *   exit
+  * }
+  * f.close(fd);
+  * @endcode
+  * 
+  * -- Notes
+  * - There should exist only one instance of this class.
+  * @todo implement simple wear levelling for file data writes.
+  * @todo implement MB_APPEND flag in open()
+  */
+
+// Configuration options.
+
+#define PAGE_SIZE 1024  // Size of page in flash.
+
+#define MAX_FD 4  // Max number of open FD's.
+
+#define FILENAME_LEN 16 // Filename length, including \0
+
+#define MBFS_LAST_PAGE_ADDR 0x3AC00 // Address of last writable page in flash.
+
+#define MAX_FILESYSTEM_PAGES 40 // Max. number of pages available.
+
+#define NO_FT_ENTRIES  10  // Number of entries in the File table  
+                            // (determines total no. files on the system)
+
+#if ( (FILENAME_LEN + MAX_FILESYSTEM_PAGES + 4 ) * NO_FT_ENTRIES ) > PAGE_SIZE
+#error NO_FT_ENTRIES is too large, file table must fit in a single page.
+#endif
+
+#if MAX_FILESYSTEM_PAGES > 127
+#error MAX_FILESYSTEM_PAGES cannot be greater than 127.
+#endif 
+
+#define MAX_FILENAME_LEN FILENAME_LEN-1
+
+#define MAGIC_WORD 0xA3E8F1C7
+
+
+// FileTableEntry_t.flags field to indicate if an FT is free/busy.
+#define FT_FREE 0x80000000
+#define FT_BUSY 0x00000000
+
+// FileTableEntry_t.flags field mask to get the file length.
+#define FT_SIZE_MASK 0x7FFFFFFF
+
+// Used in FileTableEntry_t.blocks to indicate if a block is free.
+#define FT_FREE_BLOCK_MARKER 0x80
+
+// Check if a given FileTableEntry is free.
+#define FT_IS_FREE(m) ( (m).flags & FT_FREE)
+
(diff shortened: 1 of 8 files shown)


inc/drivers/MicroBitFileSystem.h
```diff
@@ -0,0 +1,571 @@
...
+  *
+  * -- Notes
+  * - There should exist only one instance of this class.
+  * @todo implement simple wear levelling for file data writes.
+  * @todo implement MB_APPEND flag in open()
+  */
+
+// Configuration options.
+
+#define PAGE_SIZE 1024  // Size of page in flash.
```
dpgeorge: Does this need to match the erase-size of a page?  Is this the minimum amount of bytes that a given file must take up (eg a file with 1 character takes up 1024 bytes)?

finneyj:
Hi @dpgeorge 

I think there's an assumption here that a block will only hold a single file - for simplicity. Most likely based on the premise that most files will be at least 1K in size, and that we probably won't have very many at the same time.

Most of the use cases I've seen are about data logging etc. Are you thinking of use cases with different requirements? Are you thinking about storing .py files etc?

dpgeorge:
Note that 1024 will also be the size quanta (eg 1025 file takes 2 pages) so that can lead to a lot of wasted space (512 bytes on average per file).

My main question is whether this is dictated by the erase page size?  If not why not make it 512, or 256, or configurable?

> Are you thinking about storing .py files etc?

Yes we will definitely do this.  Then you can use import with your own scripts in the filesystem.

finneyj:
That would be cool. 

not sure if this is dictated by the erase size - I would presume so at present, but @alexk2k12 would be person who could confirm...

finneyj:
I guess the real tradeoff is code size against number of files and average file size... e.g. if it costs us more code space to implement the file system than we have files, then simplicity wins! But, if you expect quite a few files, then it would be worth the investment...

I guess a typical python program for microbit is likely under 1K?

inc/drivers/MicroBitFile.h
```diff
@@ -0,0 +1,198 @@
...
+      *         handle is invalid or this file was not opened in WRITE mode.
+      */
+    int append(ManagedString s);
+
+    /**
+      * Seeks to the end of the file, and appends the given character to this MicroBitFile instance.
+      *
+      * @param c The character to write to this file.
+      */
+    void operator+=(const char c);
```
pelikhan: Is this just a wrapper around ``append``? How about not support += since it does not return a proper error code and leads to potentially weird unwanted conversions.

markshannon:
If I have read the code correctly, then the following small program could destroy a page of flash.
It erases the same page 100 000 times, performing 200 000 erases in total, even though it looks like it ought not to perform more than 100 erases.
It would also be extremely slow, which might save the device but would be perplexing to the user.

```C
for (int i = 0; i < 100; i++) {
    mb_fd *fd = open("name", w);
    for (int j = 0; j < 1000; j ++) {  // Write 1k of data, 1 byte at a time.
          // Each write causes 2 page erases, 1 to the same page, 1 to a random page.
         fd->write("x", 1);
    }
}
```
Admittedly, it would take an hour to run, but it easy to imagine the inner loop occurring in a program that was run many times.

dpgeorge:
@markshannon I don't think it's the case that it erases the page just to write 1 extra byte.  Have a look at the need_erase function: it checks if the byte(s) can be written directly without the need for a temporary page and/or erasure.

markshannon:
Each write after the first (ie 99 out of the 100) will need an erase.

finneyj:
Thanks guys.

Append operations within a block should be "free" in terms of erase cycles with this approach. If Mark's sample code results in a erase cycles for each byte write then we have a bug...

@markshannon - Have you tried running your sample, or is your analysis based on code inspection?

inc/drivers/MicroBitFile.h
finneyj: Thanks @pelikhan.

Yes, this is just some syntax sugar... It's a good point that we're lacking in error codes using this interface.

@jamesadevine? How strongly do you feel about your operator overloads?

jamesadevine:
@finneyj @pelikhan 

I liked the += overloads, but I can see the drawbacks of not having any error code. 

I think supporting both of these mechanisms is a good thing, the risk of an error happening in the common case (open a file, append a few lines, close) is very low, therefore the operator overload would be the more elegant choice in terms of readability. 

Most commonly, users will probably use the .append function anyway, it's only the CPP devs that stick to the doxygen documentation that will discover it and use it! :wink: 

Having both versions does create ambiguity in behaviour, it might cause confusion and frustration... it may be better to remove the overloads.

markshannon:
@finneyj It is based on code inspection.

finneyj:
ok - thanks. I'll run it through a test harness and report back.

pelikhan:
IMO, remove it until someone complains for it. Once some feature ships it is hard to remove without breaking users, not the other way around :)

Sent from Outlook Mobile<https://aka.ms/djubve>

mmoskal:
How about using `mb_fd` directly, instead of an integer handle to it? This would reduce the RAM footprint and make the number of open files depend only on amount of RAM.

Operating systems use integer handles mostly because they live in a different address space than user programs and also need to enforce some sort of permission system. Neither is the case here.

finneyj:
OK - I've just run a test harness based on the sample from @markshannon. I essentially added some debug printf statement into MicroBitFlash::flash_write_mem to trap and report whenever a WORD was written, or a page was erased... Some interesting results. :-)

Firstly, There is no cost of an in block append, as we thought, but I see what @markshannon was getting at now. After the first run around the outer loop, of course this isn't an append operation. Subsequent writes are a random access replace operation - which is quite different, and does have an erase cost.

Now, technically, mark's sample code *doesn't* cause an erase because this code always writes the same value - and the need_erase() function picks up on this, and prevents the erase cycle. However, in the general case (i.e. if you tried to write an arbitrary value, as in the test case below), then each write operation would indeed cause an erase cycle.

```
void fs_test()
{
    uint8_t v;

    for (int i = 0; i < 100; i++) {

        SERIAL_DEBUG->printf("OPENING\n");
        int fd = fs.open("name", MB_WRITE|MB_CREAT);

        // Write 1k of data, 1 byte at a time.
        for (int j = 0; j < 1000; j++) {  
            v = (uint8_t) uBit.random(255);
            SERIAL_DEBUG->printf(".");
            fs.write(fd, &v, 1);

            // Just to save my micro:bit from excessive erase cycles in testing!
            if (i > 0)
                uBit.sleep(1000);
        }

        SERIAL_DEBUG->printf("CLOSING\n");
        fs.close(fd);
    }  
}
```

Now, unless we want to start caching blocks in RAM (which I'm pretty sure we don't, given our RAM constraints), it feels like this is 'the way it is' (tm) for random access replace operations... and at some level this is OK, as this is probably a fairly niche use-case.

HOWEVER, it also feels to me that as illustrated above, it's pretty easy for folks to do this by accident. So is perhaps is something that we should look to discourage more strongly in the API and docs. i.e. having the default position of a file opened for writing at the start of the file is perhaps not ideal...

Thoughts?

 @jamesadevine @alexk2k12 @dpgeorge @pelikhan @mmoskal

finneyj:
p.s. 

 * MicroBitFlash.cpp should also include "MicroBitConfig.h", else it can't pick up on global settings.
 * const correctness needs fixing for the write() family of operations, so a pointer to a literal string can be provided as a parameter.

dpgeorge:
If you split the filesystem data into blocks of, say, 256 bytes each (4 blocks per erase-page), then you could cache the current block in RAM and only write it out when necessary.

finneyj:
Yes, if we reduced block size that would become more feasible, probably with a single block writeback cache for each file opened for writing. We could avoid the cost if the file was read only.

This still feels a little bit like a niche case though - hate to lose RAM for something we don't typically need. Maybe creating a buffer on demand when a replace operation is attempted?

mmoskal:
Do we have a flag to delete file if it already exists? I you typically would either want to append, or create a new file, both of which avoid the problem.

finneyj:
@mmoskal - precisely, that's what I was thinking. 99.9% apps will definitely be one of those two use-cases.

I hate to speak for others, but I think @alexk2k12 was going for a POSIX compliant API, so it's READ|WRITE|CREAT. if I recall correctly, semantics of CREAT are to only create a file if it doesn't already exist. I'm not wedded to a POSIX API, but it is worth considering. This doesn't feel like the semantics we want...

note this also related to your earlier point about file descriptors...

finneyj:
Thanks to all for a good thorough review here. It sounds like we're not quite there with this yet, so here's a summary of what I think we should do next:

 * drop the append operator overloads on MicroBitFile - at least for now (sorry james!)
 * Update the MicroBiFileSystem API to use explicit WRITE_NEW / WRITE_APPEND semantics (avoiding the current default behaviour the encourages in-file replacement operations)
 * Update MicroBitFile constructor accordingly
 * fix const correctness in MicroBitFileSystem and #include MicroBitConfig.h
 * Consider introducing a configurable logical blocksize that is below the size of a flash page. I'll have a ponder on this - there are quite a few advantages here.
 * Raise another pull request

alexk2k12:
Hello all, sorry for the late reply @finneyj - been away for a week :) I'll try to address a few of the comments regarding the filesystem:

@dpgeorge - in it's current form, the page size is closely related to the erase size. I think this could be made variable, as suggested. The biggest change would be in the `remove()` method. As @finneyj said, this is a tradeoff in terms of block size vs overhead: the size of the FileTableEntry struct array would probably have to expand across multiple pages at the beginning of flash now, reducing space available for data. I will start work on making this configurable before the next PR.

@markshannon regarding your test code: that looks right. If the file already exists, and data is being overwritten, an erase will be performed on every iteration - not a test case I had designed for. A cache page in RAM would be useful here provided the page size could be reduced I think? I will start work on this now.

@mmoskal good point regarding `mb_fd`, I'll change it in the next PR.

finneyj:
Superseded by #219.

PR #142, opened 2016-05-12
Add missing const qualifier from add operator

bremoran:
This change permits friendlier use of the type such as:

```C++
ManagedString s = "A";
ManagedString b = s + ", " + 2;
```

### DIFF
--- a/inc/types/ManagedString.h
+++ b/inc/types/ManagedString.h
@@ -314,7 +314,7 @@ class ManagedString
       * display.scroll(s + p) // scrolls "abcdefgh"
       * @endcode
       */
-    ManagedString operator+ (ManagedString& s);
+    ManagedString operator+ (const ManagedString& s);
 
     /**
       * Provides a character value at a given position in the string, indexed from zero.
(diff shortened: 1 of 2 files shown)


jamesadevine:
@bremoran Thanks, I noticed this yesterday! :smile:

I will test and merge.

jamesadevine:
Thanks @bremoran - looks good!

Just a couple of thoughts based on this PR...

So now this format is valid:
```
ManagedString s = "A";
ManagedString b = s + ", " + 2;
```

Which is absolutely fantastic! 

However when you do something like this:
```
ManagedString s = "A";
ManagedString b = "abc123 " + s + " def456";
```
This doesn't work, due to operator precedence, which of course is the correct behaviour! Arguably this is confusing to a new C/C++ programmer...

A simple fix would be to precede the above example with an empty string:
```
ManagedString s = "A";
ManagedString b = ManagedString::EmptyString + "abc123 " + s + " def456";
```
However this solution lacks elegance,  is there a neater way of doing something like this?

carlosperate:
My C++ is a bit rusty, but can't you do an `std::string operator+(const std::string& slhs, const ManagedString& srhs);` overloader?

bremoran:
@carlosperate, Yes, you can do something like that. I'm working on a PR that demonstrates the technique for addition.

bremoran:
This PR should provide a solution for order of operations considerations: https://github.com/lancaster-university/microbit-dal/pull/143

PR #154, opened 2016-06-03
Added support for Just Works pairing

bluetooth-mdw:
Hopefully this is right this time :-)

### DIFF
--- a/source/bluetooth/MicroBitBLEManager.cpp
+++ b/source/bluetooth/MicroBitBLEManager.cpp
@@ -57,6 +57,22 @@ uint32_t btle_set_gatt_table_size(uint32_t size);
 
 #define MICROBIT_PAIRING_FADE_SPEED		4
 
+//
+// Local enumeration of valid security modes.  Used only to optimise pre-processor comparisons.
+//
+#define __SECURITY_MODE_ENCRYPTION_OPEN_LINK         0
+#define __SECURITY_MODE_ENCRYPTION_NO_MITM           1
+#define __SECURITY_MODE_ENCRYPTION_WITH_MITM         2
+//
+// Some Black Magic to compare the definition of our security mode in MicroBitConfig with a given parameter.
+// Required as the MicroBitConfig option is actually an mbed enum, that is not normally comparable at compile time.
+//
+
+#define __CAT(a, ...) a ## __VA_ARGS__
+#define SECURITY_MODE(x) __CAT(__,x)
+#define SECURITY_MODE_IS(x) (SECURITY_MODE(MICROBIT_BLE_SECURITY_LEVEL) == SECURITY_MODE(x))
+
+
 const char* MICROBIT_BLE_MANUFACTURER = NULL;
 const char* MICROBIT_BLE_MODEL = "BBC micro:bit";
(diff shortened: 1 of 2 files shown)


martinwork:
With this change, it looks like other values of MICROBIT_BLE_SECURITY_LEVEL (except ENCRYPTION_WITH/NO_MITM) no longer cause a call ble->securityManager().init().

bluetooth-mdw:
You're right Martin. So SECURITY_MODE_ENCRYPTION_OPEN_LINK would result in no init call. I guess the old code would have effectively called:

ble->securityManager().init(enableBonding, false, SecurityManager::IO_CAPS_DISPLAY_ONLY);

and presumably enableBonding would have been false.

I think I should incorporate this explicitly.

martinwork:
As MICROBIT_BLE_SECURITY_LEVEL is #defined ( not a static in SecurityManager), would it be more economical on compiled code to do use #if to compare MICROBIT_BLE_SECURITY_LEVEL various values and write the fixed init() call for each case?

bluetooth-mdw:
Where were you when I needed a code review earlier, Martin? :-) Excellent points. Waiting to chat with Joe before I proceed.

martinwork:
If you'd asked me earlier, I probably wouldn't have managed to come up with anything useful!

bluetooth-mdw:
Now you're just being modest :-)

profmike:
This really simplifies the pairing process, thank you!

I've been trying to find the FAQ on the following questions:
1) What's the release process and the end-user update process for features like this? 
2) When can we expect micro:bits in the field to be upgraded?

finneyj:
hi @profmike - good question! 

This will depend on the environment. So we welcome patches like this of course (great stuff!), so after a bit of peer review and due diligence testing we'll merge into our master branch, and in due course, we'll tag a release. From there, the various higher level languages and editors are free to pick it up.

Now when this happens, and how these changes are surfaced in those languages we leave up to the owners of those environments. 

 - For microbit.co.uk, the BBC are currently in code freeze to provide some stability for the kids and teachers in the UK, so any change won't take effect for a while yet. 

 - PXT tend to keep up to date, so the MSR guys usually a integrate tagged release a few days later. Any online compiles then use the new microbit-dal runtime code transparently. 

 - mbed.org is currently a manual process to update, but ARM are almost finished implementing a some magic for us that will automatically sync the mbed.org repo with our github releases.

 - Python folks ship a tailored/precanned binary version of the runtime as part of their environment, so update periodically, but not frequently

 - offline yotta/git users can update manually as they wish using yt update / git pull etc.

Hope this answers your question!

profmike:
@finneyj thank you for the detailed reply!

finneyj:
@bluetooth-mdw Just realised we never finished merging your PR in here!

There seem to be a few conflicts now - can you please take a look so that we can get JW pairing merged?

bluetooth-mdw:
@finneyj will do.

bluetooth-mdw:
You can kill this PR. Resubmitted as https://github.com/lancaster-university/microbit-dal/pull/216

finneyj:
ok

Issue #156, opened 2016-06-03
Rollup optional pull specification into getDigitalValue


cefn:
This is a change request following some consensus on medium-term workarounds for the surprising pull-down defaults of digitalWrite (surprising for experienced makers anyway). 

This relates to the discussion at... https://github.com/bbcmicrobit/micropython/issues/288 

The request is to provide a signature for getDigitalValue which promotes visibility of software-controllable pull behaviour and makes the existing default pull-setting visible and more accessible, as well as seamlessly allowing getDigitalValue calls to be opinionated about pull, and encouraging the downstream languages to extend their own mappings of the getDigitalValue call to include pull specification.

THE PROPOSAL 

They should be set by...

* calling getDigitalValue with an explicit pull argument (this is the extra requested signature, of which the below is a special case of resorting to a default value for this argument)
* calling getDigitalValue with no explicit pull at which point they are implicitly set to e.g.  MICROBIT_PULL_DEFAULT which is currently pull-down, though there should perhaps be a separate discussion about this. If previously set, getDigitalValue without an explicit pull will not overwrite the value. 
* explicitly pulling a pin (but not as part of a getDigitalValue call).

An alternative to avoid the potential for getDigitalValue() to be mistakenly substituted for setDigitalValue() which would end up with the same argument signature and a single character difference, is to provide getPulledValue() but this does not have the benefit of encouraging downstream languages to track an existing API they are bound to, and they may remain oblivious to the need to expose this key operation.

finneyj:
Thanks @cefn. Agree with all that, except the last point. getDigitalValue() is the known name - I think we should stuck with it.

I'll draft up a pull request (pun intended!) for review.

finneyj:
https://github.com/lancaster-university/microbit-dal/pull/158

finneyj:
@cefn @dpgeorge @pelikhan 

FYI, as requested. https://github.com/lancaster-university/microbit-dal/pull/158

PR #158, opened 2016-06-03
microbit: Added getDigitalValue overload for PullMode #156

finneyj:
- Introduced an overload to MicroBitIOPin::getDigitalValue() to permit the
setting of a specific pull mode at the time of reading.

 - Bugfix of MicroBitIOPin::setPull() to persist preferred pull settings

 - Added configuration options to allow the default PullMode to be set via
   compile time option through MicroBitConfig.h or YOTTA_CONFIG


inc/drivers/MicroBitPin.h
```diff
@@ -95,8 +95,8 @@ class MicroBitPin : public MicroBitComponent
 {
     // The mbed object looking after this pin at any point in time (untyped due to dynamic behaviour).
     void *pin;
-
     PinCapability capability;
+    uint8_t pullMode;
```
dpgeorge: This is going to add (probably) 4 bytes of RAM to each pin.  That means taking about 80 bytes when all ~20 pins are created.

finneyj: checked that - fits within current word alignment bounds, so adds no additional overhead.

dpgeorge: Phew!

finneyj: :-)

finneyj:
Merging early as @pelikhan has kindly agreed to test on PXT if we can get it into master as part of this week's planned tagged interim release. Any further comments will still be welcomed and considered.

Issue #175, opened 2016-06-28
Radio receive - event queue jams due to unprotected queueDepth decrement


whs-che:
MicroBitRadio.cpp

Variable queueDepth is manipulated in the radio ISR , via queueRxBuf(), and by an idle thread. This can cause the queueDepth to decrement to 255 and hence break the max buffers detection logic. 

i.e. idle thread reads current value of queueDepth  -> interrupt routine gets in and successfully increments the count -> idle thread resumes and promptly overwrites the correct value. Since there are now "excess" packets in the queue, subsequent decrements drive the counter to 255. Since this is larger than MICROBIT_RADIO_MAXIMUM_RX_BUFFERS no more queuing can take place. 
A possible method of solution is shown below and appears to work and also encompasses the list manipulation just to be safe. Without more in-depth knowledge of the DAL I can't comment on potential side-effects with this fix. Is there a better way to serialise access here?

Failure on test system (7 microbits broadcasting at 100ms) occurs within 30mins. Current test with the fix is at 36 hours and still okay.

```
FrameBuffer* MicroBitRadio::recv()
{
    FrameBuffer *p = rxQueue;

    if (p)
    {
        NVIC_DisableIRQ(RADIO_IRQn);  // Added - protection from isr

        rxQueue = rxQueue->next;
        queueDepth--;

        NVIC_EnableIRQ(RADIO_IRQn); // Added - re-enable isr

    }

    return p;
}
```

jamesadevine:
@whs-che Great catch! 

That also looks like a suitable patch :smile:

Is there any way you could form this into a pull request?

finneyj:
Hi @whs-che

Yes, a classic race condition that sneaked in. Thanks for the diligent testing! 

Agree this is a suitable fix. Technically NVIC_DisableIRQ isn't always safe (if soft device is running, there's a n alternative wrapper function), but as Radio and SoftDevice are currently mutually exclusive, we're fine here I think.

Let's wrap it up in a Pull Request and merge it.

jamesadevine:
Fixed by PR #176

PR #176, opened 2016-06-30
Protection from ISR added to queueDepth counter (#175 )

whs-che:
MicroBitRadio.cpp

Variable queueDepth is manipulated in the radio ISR , via queueRxBuf(), and by an idle thread. This can cause the queueDepth to decrement to 255 and hence break the max buffers detection logic.

i.e. idle thread reads current value of queueDepth -> interrupt routine gets in and successfully increments the count -> idle thread resumes and promptly overwrites the correct value. Since there are now "excess" packets in the queue, subsequent decrements drive the counter to 255. Since this is larger than MICROBIT_RADIO_MAXIMUM_RX_BUFFERS no more queuing can take place. 
A possible method of solution is shown below and appears to work and also encompasses the list manipulation just to be safe. Without more in-depth knowledge of the DAL I can't comment on potential side-effects with this fix. Is there a better way to serialise access here?

Failure on test system (7 microbits broadcasting at 100ms) occurs within 30mins. Current test with the fix is at 36 hours and still okay.

```
FrameBuffer* MicroBitRadio::recv()
{
    FrameBuffer *p = rxQueue;

    if (p)
    {
        NVIC_DisableIRQ(RADIO_IRQn);  // Added - protection from isr

        rxQueue = rxQueue->next;
        queueDepth--;

        NVIC_EnableIRQ(RADIO_IRQn); // Added - re-enable isr

    }

    return p;
}
```

### DIFF
--- a/source/drivers/MicroBitRadio.cpp
+++ b/source/drivers/MicroBitRadio.cpp
@@ -452,8 +452,14 @@ FrameBuffer* MicroBitRadio::recv()
 
     if (p)
     {
+         // Protect shared resource from ISR activity
+        NVIC_DisableIRQ(RADIO_IRQn); 
+
         rxQueue = rxQueue->next;
         queueDepth--;
+
+        // Allow ISR access to shared resource
+        NVIC_EnableIRQ(RADIO_IRQn);
     }
 
     return p;


jamesadevine:
Thanks very much @whs-che ! :smile:

PR #207, opened 2016-08-30
Make sure there is only one empty string

mmoskal:
This makes sure we always re-use a single empty string instance. We were not doing that in some extreme circumstances


jamesadevine:
Ok, that was not so interesting... 😛

mmoskal:
Thanks!

PR #209, opened 2016-08-31
PWM Control characteristic added to IO Pin Service

bluetooth-mdw:
...and a typo fixed. No extra charge for that ;-)


bluetooth-mdw:
btw tested with an LED using the following data:

```
![image](https://cloud.githubusercontent.com/assets/8341666/18126738/84b131b8-6f74-11e6-904b-aa3aeaa1b5f4.png)

```

Not tested with a buzzer and varying frequency. Need someone else to do that.

PR #214, opened 2016-09-26
Added physical web transmission

Thomas101:
This adds two methods to the `MicroBitBLEManager`. The first one `advertisePhysicalWebUrl` allows you transmit a physical web url from the microbit. The second `stopAdvertise` stops all outgoing ble advertisements


jamesadevine:
@Thomas101 This is great, a couple of comments:

* Can we have a ManagedString variant of `advertisePhysicalWebUrl`?
* A default for `calibratedPower`
* I think `stopAdvertise` should be renamed to `stopAdvertising`?

Thomas101:
Hi @jamesadevine I've added a ManagedString variant and renamed `stopAdvertise` to `stopAdvertising`.

As regards the `calibratedPower` the value provided here depends on two things, firstly the transmission power that's set on the microbit and secondly the environment the microbit is placed within (e.g. a case, no case, behind a brick etc). To calculate this value you should measure the received signal strength from the microbit with your desired configuration and then re-supply this to the function call which will provide accurate distance measurements. I've added some more docs around the function to reflect this and also linked to the physical web spec which describes this in more detail.

I'm happy to take more direction if you still think it needs a default value or maybe a different function completely that uses a wildcard configuration - giving an incorrect distance measurement but still transmitting correctly.

finneyj:
Thanks @Thomas101 - looking good. Just a few thoughts:

- I agree we should keep the calibratedPower variable as an explicitly defined parameter. This is something the application programmer should really define. As you say, it is affected by so many things.

- I'm a bit puzzled by the type of the calibratedPower parameter though. Isn't the value a signed 8 bit int? As defined below? If so, could we update this field to be a int8_t, which would make it much cleaner for the application programmer to specify the correct value dbm value...

https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.characteristic.tx_power_level.xml&u=org.bluetooth.characteristic.tx_power_level.xml

- Maybe an optional parameter that also defines the transmission interval? 400ms is a great default, but there may be some apps where the user might want to know quickly if a beacon goes out of range etc?

martinwork:
Should this be placed in it's own source file (and class?) rather than adding to MicroBitBLEManager? A quick test seemed to show that adding an initialised but unreferenced global const char * to MicroBitBLEManager.cpp adds more than it's size to the compiled firmware and an unreferenced function adds a few bytes. I am assuming that an unreferenced source file won't add to the product.

remay:
I suspect that separating advertising from the BLEManager would be a useful thing to do.

finneyj:
That would be ideal, but perhaps beyond the scope of this patch? (@Thomas101 - I'd welcome such a patch, but totally understand if you want to scope this out). I am a little surprised the compiler isn't optimizing this out though... @martinwork was that a gcc/yotta build?

martinwork:
yotta with target bbc-microbit-classic-gcc

Thomas101:
@finneyj Thanks, I've changed the calibration to use `int8_t` as recommended and also added the advertising interval.

@martinwork I've wrapped the physicalweb code around an ifdef which can be configured from the `config.json` using the `microbit-dal.bluetooth.physical_web` key as per the other bluetooth services. This should mean that there is no impact on binary size if the physical web is not enabled.

If the `config.json` fix is alright are there any docs that will need updating to reflect this?

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -197,6 +199,28 @@ class MicroBitBLEManager : MicroBitComponent
...
+        * More information can be found at https://github.com/google/eddystone/tree/master/eddystone-url#tx-power-level
+        * @param internval: the advertising interval of the beacon
+	*/
+    	void advertisePhysicalWebUrl(char* url, int8_t calibratedPower, uint16_t interval = MICROBIT_BLE_PW_ADV_INTERVAL);
+
+        /**
+        * Transmits a physical web url, but accepts a ManagedString as a url. For more info see
+        * advertisePhysicalWebUrl(char* url, int8_t calibratedPower, uint16_t interval)
+        */
+    	void advertisePhysicalWebUrl(ManagedString url, int8_t calibratedPower, uint16_t internval = MICROBIT_BLE_PW_ADV_INTERVAL);
```
cpseager: internval typo

martinwork:
@Thomas101 good thinking with ifdef!

Thomas101:
:-) If you're looking for an example usage of the code, I've put one up [here](https://github.com/showio/microbit-physicalweb)

jamesadevine:
@Thomas101 Great thinking with an ifdef :smile:

I'm wondering if we should perhaps adopt a common mechanism for beacon services through class instantiation as @martinwork suggested. It would also mean that the user is in control of when the class is compiled and used and would also future proof against cases where users wanted a different kind of beacon service.

I'm up for a discussion on this one @Thomas101, what are your thoughts?

As an aside: Can beacon data be the only advertisement data that is available? Could we also preserve the advertisement data for a micro:bit?

Thomas101:
@jamesadevine I think there's definite value in adopting a common mechanism for different beacon services in the future.

For now our focus is on the Physical Web because there are some really ace demos that schools can build using the physical web and web bluetooth using the latest versions of chrome which we're going to be writing some blog posts and tutorials about. See a couple of early video deomos [here](https://www.youtube.com/watch?v=DCri83kUS7M) and [here](https://www.youtube.com/watch?v=xVvP1gr5SOI)

bluetooth-mdw:
Late to the party here....

I notice the advertising used here is of the connectable type. 

```
ble->setAdvertisingType(GapAdvertisingParams::ADV_CONNECTABLE_UNDIRECTED);
```
Is this intentional and if so what is the thinking here? A beacon would normally advertising as a non-connectable device (i.e. a GAP Broadcaster). Physical Web does define a GATT Configuration Service and if that was implemented, then when in "configuration mode", the beacon would switch to using connectable advertising packets but otherwise would use non-connectable.  The Eddystone spec says "We recommend that your beacons spend the majority of their lives in a non-connectable state,....."

I also noticed this comment:

bluetooth-mdw:
p.s. I agree a more generalised beacon framework which supports iBeacon, AltBeacon, Eddystone and whatever other beacony stuff people want is the way to go in the future, but getting Eddystone available in a simple form now, is a good step.

What I would suggest though is that the Eddystone support we have here, which is in fact only partial support for Eddystone, perhaps be refactored a little to anticipate adding the other Eddystone frame types? URL is only one of 4 frame types.

Final comment: I'd prefer to see EddyStone in the comments and method names etc rather than PhysicalWeb. PhysicalWeb is the name of a project. EddyStone is the protocol. May as well educate people to use the right terminology :-)

martinwork:
@jamesadevine Thanks for giving me somewhat more credit than I deserve!
My concern was mostly to avoid adding unused code and data to everyone's builds. It also occurred to me that MicroBitBLEManager doesn't need these functions and data to be part of it's class and they don't need to be in MicroBitBLEManager to do their job. So it would make sense for them to be separate - more so if this is a first step on the way to something bigger.

Thomas101:
@bluetooth-mdw agree with the correct naming of eddystoneUrl - its been changed to reflect this, and therefore allows other eddystone types in the future with eddystoneEid, eddystoneTlm etc.

Regarding the ADV_CONNECTABLE_UNDIRECTED, I can imagine plenty of use-cases in schools where the microbit has much more use when it stays connectable. For example if broadcasting a page and expecting another device to connect over web bluetooth to interact. Would it be best to make this a config option?

jamesadevine:
@martinwork Any ideas as to why it's not being optimised out?

@Thomas101 @bluetooth-mdw I think the primitive URL frame is great for now! We should merge with the intention of breaking these out into separate beacon classes. 

Perhaps in the future we could overload these functions with the different Beacon protocols, and it will internally instantiate the appropriate class.

finneyj:
@Thomas101.  Thanks for the quick turnaround - all looks good to me. Would be great if you could update the docs around here to record the new yotta option: https://lancaster-university.github.io/microbit-docs/advanced/#compile-time-options-with-microbitconfigh

The repo containing the content is here:
https://github.com/lancaster-university/microbit-docs

Agree we should look at other services in due course, but we can always refactor a bit when they come along. Let's merge this in.

bluetooth-mdw:
@Thomas101 I think by default, if using Eddystone, advertising packets should indicate non-connectable. By default, none of the GATT services should be instantiated if the micro:bit is being used as a EddyStone beacon either. If it's not connectable, those services have no use.

If any GATT services are included with the EddyStone URL use though, this could/should result in connectable advertising being used automatically.

Finally (?), it would be useful to be able to dynamically change the ADV type by a function call. Beacons sometimes let you switch them into config or maintenance mode so that you can then connect to them and configure (e.g.) their new URL over the air.

That's my wish list :-)

Thomas101:
I've opened a PR in the docs https://github.com/lancaster-university/microbit-docs/pull/36

martinwork:
@jamesadevine I'm afraid I'm no gcc guru, but from searching around...
Maybe it's something to do with file vs function level linking. This is the most interesting reference I could find:
http://stackoverflow.com/questions/4274804/query-on-ffunction-section-fdata-sections-options-of-gcc

finneyj:
All those options are already enabled on the mbed yotta target... which is why I was surprised that this isn't compiled out. curious.

finneyj:
I think we've all reached consensus here, so merging in. This will go into master, and be tagged in the next release. Thanks for your contribution @Thomas101!

p.s. I'd also welcome a PR adding a simple beacon sample program based on your API in the microbit-samples repo... I can see this being a pretty popular use case, as a sample might accelerate uptake of your work?

Thomas101:
Thanks @finneyj we've got an example over at [showio/microbit-physicalweb](https://github.com/showio/microbit-physicalweb) but once the next tag comes out I'll add it into the samples repo :). Do you have any schedule on this?

Thanks again everyone!

bluetooth-mdw:
@finneyj Did we? My comments about advertising have been ignored I think. Still looks like we're using connectable advertising by default. I'm OK with this being something you can enable by parameter to advertiseEddystoneUrl() but it should not use connectable advertising by default. This is not how beacons generally work and if we're giving kids an intro to beacons it's a good idea to do it in a way which mirrors the real world. Yes there will be times where you want it to be connectable too. If there's a good reason to default to connectable let's hear it..... DFU for example?

+    ble->setAdvertisingType(GapAdvertisingParams::ADV_CONNECTABLE_UNDIRECTED);

Thomas101:
I'll put a new pull request that adds this a parameter. This can then address both use cases of using the microbit as an eddystone beacon and also using the microbit as a web bluetooth controlled device that also transmits its controlling webpage on the physical web.

finneyj:
@bluetooth-mdw - apologies I thought we were parking that one for later. merging in this additional patch.

bluetooth-mdw:
@finneyj You are forgiven ;-)

finneyj:
phew. ;-)

@Thomas101 we'll be merging in the local file system branch soon (PR probably going in today). We'll tag a release once that's merged.

martinwork:
@finneyj @jamesadevine  About data-sections etc... Did you confirm what I reported? Maybe I'm doing it wrong! My yotta build is on OS X, in case that makes a difference.

It seems I was mistaken about unused functions. It's all about the data. The function I tested with contained a string constant. Maybe the following can shed some light... tested by adding to main, so it's nothing to do with linking from libraries. It looks like the build is removing unused functions but failing to remove data that is only used by those functions or not used at all.

Adding an unused function doesn't affect the hex
void unused_function() { uBit.display.clear(); }

Constants in an unused function add to the hex:
void unused_function() { uBit.display.printAsync( "unused"); }

Adding static lets the compiler filter out unused constants with a warning:
const char *unused = "a very long string";
static const char *unused = "a very long string";

A reference to an otherwise unused string in an unused function, brings back the size increase:
static const char *unused = "a very long string";
void unused_function() { uBit.display.printAsync( unused); }

Adding more code doesn't change the size:
static const char *unused = "a very long string";
void unused_function() { uBit.display.printAsync( unused); uBit.display.clear(); }

PR #219, opened 2016-10-03
Microbit mbfs

finneyj:
Revised MicroBitFileSystem implementation.
 
This PR introduces a variation of @alexk2k12's posix file system implementation to address weaknesses noted in #139. It should be noted that a significant part of this contribution was driven by @ornberg and @AHerczeg (thanks guys!).

This patch introduces the following changes over #139.

 - Introduction of logical blocks of a configurable size (any power of 2 fraction of the device's PAGESIZE - 1K for nrf51822).
 - A move to a non-contiguous logical block tracking table.
 - Round robin block allocation policy to improve load balancing.
 - introduction of a configurable in memory writeback cache to reduce/mitigate impact of frequent overwrite operations within a single file.
 - Optimisations around all operations to minimise the number of erase cycles of FLASH pages.
 - removal of '+=' overload operators from MicroBitFile

### DIFF
--- a/inc/drivers/MicroBitFile.h
+++ b/inc/drivers/MicroBitFile.h
@@ -1,0 +1,195 @@
+#ifndef MICROBIT_FILE_H
+#define MICROBIT_FILE_H
+
+#include "mbed.h"
+#include "MicroBitConfig.h"
+#include "MicroBitFileSystem.h"
+#include "ManagedString.h"
+
+#define READ                            MB_READ
+#define WRITE                           MB_WRITE
+#define READ_AND_WRITE                  READ | WRITE
+#define CREATE                          MB_CREAT
+
+class MicroBitFile
+{
+
+    int fileHandle;
+    ManagedString fileName;
+
+    public:
+
+    /**
+      * Creates an instance of a MicroBitFile and creates a new file if required.
+      *
+      * @param fileName the name of the file to create/open.
+      *
+      * @param mode One of: READ, WRITE, READ_AND_WRITE. Defaults to READ_AND_WRITE.
+      */
+    MicroBitFile(ManagedString fileName, int mode = READ_AND_WRITE);
+
+    /**
+      * Seeks to a position in this MicroBitFile instance from the beginning of the file.
+      *
+      * @param offset the number of bytes to seek, relative to the beginning of the file.
+      *
+      * @return the new seek position, MICROBIT_NOT_SUPPORTED if the current file handle is invalid,
+      *         MICROBIT_INVALID_PARAMETER if the given offset is negative.
+      */
+    int setPosition(int position);
+
+    /**
+      * Returns the current position of the seek head for the current file.
+      *
+      * @return the new seek position, MICROBIT_NOT_SUPPORTED if the current file handle is invalid.
+      */
+    int getPosition();
+
+    /**
+      * Writes the given bytes to this MicroBitFile instance at the current position.
+      *
+      * @param bytes a pointer to the bytes to write to this file.
+      *
+      * @param len the number of bytes to write.
+      *
+      * @return The number of bytes written, MICROBIT_NOT_SUPPORTED if the current file
+      *         handle is invalid, MICROBIT_INVALID_PARAMETER if bytes is invalid, or
+      *         len is negative.
+      */
+    int write(const char *bytes, int len);
+
+    /**
+      * Writes the given ManagedString to this MicroBitFile instance at the current position.
+      *
+      * @param s The ManagedString to write to this file.
+      *
+      * @return The number of bytes written, MICROBIT_NOT_SUPPORTED if the current file
+      *         handle is invalid, MICROBIT_INVALID_PARAMETER if bytes is invalid, or
+      *         len is negative.
+      */
+    int write(ManagedString s);
+
+    /**
+      * Reads a single character from the file at the current position.
+      *
+      * @return the character, or MICROBIT_NOT_SUPPORTED if the current file handle
+      *         is invalid.
+      */
+    int read();
+
+    /**
+      * Reads from the file into a given buffer.
+      *
+      * @param buffer a pointer to the buffer where data will be stored.
+      *
+      * @param size the number of bytes that can be safely stored in the buffer.
+      *
+      * @return the number of bytes read, or MICROBIT_INVALID_PARAMETER if buffer is
+      *         invalid, or the size given is less than 0.
+      */
+    int read(char *buffer, int size);
+
+    /**
+      * Reads from the current MicroBitFile into a given buffer.
+      *
+      * @param size the number of bytes to be read from the file.
+      *
+      * @return a ManagedString containing the requested bytes, oran empty ManagedString
+      *         on error.
+      */
+    ManagedString read(int size);
+
+    /**
+      * Removes this MicroBitFile from the MicroBitFileSystem.
+      *
+      * @return MICROBIT_OK on success, MICROBIT_INVALID_PARAMETER if the given filename
+      *         does not exist, MICROBIT_CANCELLED if something went wrong.
+      */
+    int remove();
+
+    /**
+      * Seeks to the end of the file, and appends the given ManagedString to this MicroBitFile instance.
+      *
+      * @param bytes a pointer to the bytes to write to this file.
+      *
+      * @param len the number of bytes to write.
+      *
+      * @return The number of bytes written, MICROBIT_NOT_SUPPORTED if the current file
+      *         handle is invalid or this file was not opened in WRITE mode.
+      */
+    int append(const char *bytes, int len);
+
+    /**
+      * Seeks to the end of the file, and appends the given ManagedString to this MicroBitFile instance.
+      *
+      * @param s The ManagedString to write to this file.
+      *
+      * @return The number of bytes written, MICROBIT_NOT_SUPPORTED if the current file
+      *         handle is invalid or this file was not opened in WRITE mode.
+      */
+    int append(ManagedString s);
+
+    /**
+      * Seeks to the end of the file, and appends the given character to this MicroBitFile instance.
+      *
+      * @param c The character to write to this file.
+      */
+    void operator+=(const char c);
+
+    /**
+      * Seeks to the end of the file, and appends the given sequence of characters to this MicroBitFile instance.
+      *
+      * @param s The sequence of characters to write to this file.
+      *
+      */
+    void operator+=(const char* s);
+
+    /**
+      * Seeks to the end of the file, and appends the given ManagedString to this MicroBitFile instance.
+      *
+      * @param s The ManagedString to write to this file.
+      *
+      */
+    void operator+=(ManagedString& s);
+
+    /**
+      * Returns the handle used by this MicroBitFile instance.
+      *
+      * @note This member function will also inform the user of any errors encountered
+      *       during the opening of this MicroBitFile. At open, the handle is set
+      *       to the return value from MicroBitFileSystem.open().
+      */
+    int getHandle();
+
+    /**
+      * Opens this MicroBitFile instance if the file has previously been closed.
+      *
+      * @param mode One of: READ, WRITE, READ_AND_WRITE. Defaults to READ_AND_WRITE.
+      *
+      * @return MICROBIT_OK on success, MICROBIT_NOT_SUPPORTED if the file is already open,
+      *         MICROBIT_INVALID_PARAMETER if the filename is too large, MICROBIT_NO_RESOURCES
+      *         if the file system is full, or memory could not be allocated..
+      *
+      * @note MicroBitFiles are opened at construction and are implicitly closed at
+      *       destruction. They can be closed explicitly using the close() member function.
+      */
+    int open(int mode = READ_AND_WRITE);
+
+    /**
+      * Closes this MicroBitFile instance
+      *
+      * @return MICROBIT_OK on success, MICROBIT_NOT_SUPPORTED if the file handle
+      *         is invalid.
+      *
+      * @note MicroBitFiles are opened at construction and are implicitly closed at
+      *       destruction. They can be closed explicitly using this member function.
+      */
+    int close();
+
+    /**
+      * Destructor for MicroBitFile. Implicitly closes the current file.
+      */
+    ~MicroBitFile();
+};
+
+#endif
--- a/inc/drivers/MicroBitFile.h
+++ b/inc/drivers/MicroBitFile.h
@@ -26,7 +26,7 @@ class MicroBitFile
       *
       * @param mode One of: READ, WRITE, READ_AND_WRITE. Defaults to READ_AND_WRITE.
       */
-    MicroBitFile(ManagedString fileName, int mode = READ_AND_WRITE);
+    MicroBitFile(ManagedString fileName, int mode = READ | WRITE | CREATE);
 
     /**
       * Seeks to a position in this MicroBitFile instance from the beginning of the file.
@@ -128,6 +128,14 @@ class MicroBitFile
       *         handle is invalid or this file was not opened in WRITE mode.
       */
     int append(ManagedString s);
+
+    /**
+     * Determines if this MicroBitFile instance refers to a valid, open file.
+     *
+     * @return true if this file is valid, false otherwise.
+     *
+     */
+    bool isValid();
 
     /**
       * Returns the handle used by this MicroBitFile instance.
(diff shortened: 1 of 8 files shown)


source/drivers/MicroBitFileSystem.cpp
```diff
@@ -0,0 +1,1278 @@
...
+    id = 0;
+    
+    while (file && dirent)
+    {
+        if (file->dirent == dirent)
+            return MICROBIT_NOT_SUPPORTED;
+
+        if (file->id == id)
+        {
+            id++;
```
remay: Shjould there be a chack for id > MAX supported open files.  I know that we're memory constrained, but in theory we could overlow id resulting in a negative return value.

inc/drivers/MicroBitFile.h
```diff
@@ -0,0 +1,172 @@
...
+      *
+      * @param bytes a pointer to the bytes to write to this file.
+      *
+      * @param len the number of bytes to write.
+      *
+      * @return The number of bytes written, MICROBIT_NOT_SUPPORTED if the current file
+      *         handle is invalid, MICROBIT_INVALID_PARAMETER if bytes is invalid, or
+      *         len is negative.
+      */
+    int write(const char *bytes, int len);
```
remay: Do we need a MicroBitFile::Flush(), given that small writes can be cached?  Or are we trying to discourage it?  It guess it could be achieved by doing a close() immediately followed by an open().  While developing code it's sometimes useful to ensure your data got persisted before doing something that might crash ...

source/drivers/MicroBitFile.cpp
```diff
@@ -0,0 +1,280 @@
...
+    this->fileName = fileName;
+
+    MicroBitFileSystem* fs;
+
+    if(MicroBitFileSystem::defaultFileSystem == NULL)
+        fs = new MicroBitFileSystem();
+    else
+        fs = MicroBitFileSystem::defaultFileSystem;
+
+    fileHandle = fs->open(fileName.toCharArray(), mode | CREATE);
```
remay: Also want to comment about the explicit adding of CREATE to mode here - make it impossible for me to test for the existence of a file.  But perhaps I should be using the MicroBitFileSystem directly to do that?

remay: Perhaps only add CREATE if the file is going to be writable?  READ | CREATE doesn't make much sense to me.   But still leaves the problem of how to signal failure from the constructor.

jamesadevine: @remay `fs->open` will return an invalid file handle, so subsequent operations will fail on the instance.

I don't think the `CREATE` behaviour has any horrible side affects, although I agree that it is a bit weird to create a file if you only want to read it, so that should change. Also it should be noted that `mode` defaults to `READ_AND_WRITE` which is suitable for the general audience who will be using this class over `MicroBitFileSystem`, I think the `CREATE` behaviour should remain implicit with the exception of `READ`. 

Two potential solutions:
* Have an exception case for `READ`.
* Change the default of `mode` to be `READ_AND_WRITE | CREATE`.

I think I prefer the second solution, that way any modifications to that `mode` parameter will be as the user expects.

remay: @jamesadevine thank you.  I like the second option too - a sensible default, but can be explicitly changed by the caller to get the exact behaviour desired.  Probably want the same change to the default mode and implementation on open().

remay: The issue with failure in the constructor is more tricky.  I don't t like the "action at a distance" problem where a subsequent file operation fails (with MICROBIT_NOT_SUPPORTED) is the open() nin the constructor failed.   I definitely don't like having to call getfilehandle() to determine the error state of the constructor.  I'm not sure what the right answer is though.

source/drivers/MicroBitFile.cpp
```diff
@@ -0,0 +1,280 @@
...
+  *
+  * @note MicroBitFiles are opened at construction and are implicitly closed at
+  *       destruction. They can be closed explicitly using the close() member function.
+  */
+int MicroBitFile::open(int mode)
+{
+    if(fileHandle >= 0)
+        return MICROBIT_NOT_SUPPORTED;
+
+    int ret = MicroBitFileSystem::defaultFileSystem->open(fileName.toCharArray(), mode | CREATE);
```
remay: MIght be good for a method without the mode parameterthat uses the original mode  ..  so that doing f->close(); f->open(); works?

inc/drivers/MicroBitFileSystem.h
```diff
@@ -0,0 +1,468 @@
...
+    FileDescriptor *openFiles = NULL;
+
+    /**
+      * Initialize the flash storage system
+      *
+      * The file system is located dynamically, based on where the program code
+      * and code data finishes. This avoids having to allocate a fixed flash
+      * region for builds even without MicroBitFileSystem. 
+      *      
+      * This method checks if the file system already exists, and loads it it.
```
finneyj: Thanks - corrected.

inc/drivers/MicroBitFileSystem.h
```diff
@@ -0,0 +1,468 @@
...
+    */
+    DirectoryEntry* createDirectoryEntry(DirectoryEntry *directory);
+
+    /**
+    * Refresh the physical page associated with the given block.
+    * Any logical blocks marked for deletion on that page are recycled.
+    *
+    * @param block the block to recycle.
+    * @param type One of MBFS_BLOCK_TYPE_FILE, MBFS_BLOCK_TYPE_DIRECTORY, MBFS_BLOCK_TYPE_FILETABLE. 
+    * Erases and regenerates the given block, recycling and data marked for deletion.
```
finneyj: fixed.

source/drivers/MicroBitFileSystem.cpp
```diff
@@ -0,0 +1,1278 @@
...
+/**
+  * Retrieve the next block in a chain.
+  *
+  * @param block A valid block number.
+  *
+  * @return The block number of the next block in the file.
+  */
+uint16_t MicroBitFileSystem::getNextFileBlock(uint16_t block)
+{
+    return fileSystemTable[block];
```
finneyj: :-) nope. The file table refers to the next block in a chain, so fileSystemTable[block] refers to the next block in sequence... so we're good here. This is the same principle as FAT really, just a bit leaner and optimised for RAM and memory mapped FLASH... (perhaps we should call it "THIN"!)

source/drivers/MicroBitFileSystem.cpp
finneyj: We don't really have a max supported open files anymore - we're just limited by available RAM. Allocated fds always count from zero and recycle from zero too (so there's no fragmentation of the ID space - the lowest available fd number will always be used). this means that we'd need to have 2^31 OPEN files (~2 billion) to overflow into a negative id... so I think we're safe! :-)

inc/drivers/MicroBitFile.h
finneyj: hmm... that's an interesting point. A bit like the UNIX sync() operation. i remember always doing sync(); sync(); sync();  before testing any device drivers I wrote while I was an undergraduate. :-)

I had assumed that close() would be used for this, because we need to ensure all state is updated (cache flushed, directory entries updated, etc). But perhaps there is also a need for an explicit flush() operation that keeps the file open...

source/drivers/MicroBitFile.cpp
```diff
@@ -0,0 +1,280 @@
...
+    return append(s.toCharArray(), s.length());
+}
+
+/**
+  * Returns the handle used by this MicroBitFile instance.
+  *
+  * @note This member function will also inform the user of any errors encountered
+  *       during the opening of this MicroBitFile. At open, the handle is set
+  *       to the return value from MicroBitFileSystem.open().
+  */
```
finneyj: as above... such an approach would help here too?

source/drivers/MicroBitFile.cpp
finneyj: Yes - docs should make this clear. 

not quite sure what you mean by "without the mode parameter that uses the original mode"... Do you mean caching the last used file mode? This could be tricky, as the MicroBitFile object will have gone out of scope... and we don't want to have static data as we could have multiple files accessed... I'm also not sure this would really help to improve clarity (holding state like this can often make things more complex).

inc/drivers/MicroBitFile.h
finneyj: I've just drafted this up - it's a very clean patch with minimal overhead, so I'll test it through and commit it...

source/drivers/MicroBitFileSystem.cpp
```diff
@@ -0,0 +1,1278 @@
...
+    // If we have a zero length, then dynamically determine our geometry.
+    if (flashStart == 0)
+    {
+        // Flash start is on the first page after the programmed ROM contents.
+        // This is: __etext (program code) + static data.
+        // Size of static data is calculated from __data_end__ and __data_start__
+        // (See the linker script)
+
+        flashStart = (uint32_t)&__etext + ((uint32_t)&__data_end__ - (uint32_t)&__data_start__);
+        flashStart = ((uint32_t)flashStart & ~(PAGE_SIZE-1)) + PAGE_SIZE;
```
finneyj: ARMCC has different symbols: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0474i/CHDBIJJD.html

so some conditional compilation required here to support the mbed.org online compiler. Good catch @remay.

inc/drivers/MicroBitFile.h
remay: Thanks - interested to see the implementation.  I had missed the bit about updating the other state.  I assume it's the equivalent of f.close(); followed y f.open(original_mode)?  (But haven't digested your response to my comments on open() yet).

source/drivers/MicroBitFile.cpp
remay: Given that we want to avoid exceptions, C++ doesn't give us any nice way of dealing with this.  I'd be happy with an is_valid() or equivalent method that checks the filehandle value and returns a boolean, allowing

```cpp
MicroBitFile f("hello.txt");
if (!f.is_valid()) {
   // handle error
}

File: source/drivers/MicroBitFile.cpp (hunk 8 above)
Comment (remay) [inline source/drivers/MicroBitFile.cpp]: Really this was the same comment as above, so let's leave the discussion above.

File: source/drivers/MicroBitFile.cpp (hunk 4 above)
Comment (remay) [inline source/drivers/MicroBitFile.cpp]: mode: see my comments regarding this on the constructor.  The current implementation doesn't allow me to re-open a file in READ-only mode, with an error if it doesn't exist.  I maybe don't understand correctly the intended purpose of open().

I was thinking about caching the original mode in the MicroBitFile object.  I can't call open() at all if the object has gone out of scope, so it can only be called if it is still in scope ... my intention had really been to allow
```cpp
MicroBitFile f("hello.txt");
// write some stuff to the file
f.close();
f.open();
// file is open and in the same state (bar the seek position) as before the close(), but flushed to flash and with all state guaranteed to be updated.

File: source/drivers/MicroBitFileSystem.cpp (hunk 9 above)
Comment (remay) [inline source/drivers/MicroBitFileSystem.cpp]: :-)

File: source/drivers/MicroBitFileSystem.cpp (hunk 7 above)
Comment (remay) [inline source/drivers/MicroBitFileSystem.cpp]: Thank you.  Just looked odd when scanning.

File: source/drivers/MicroBitFileSystem.cpp (hunk 1 above)
Comment (remay) [inline source/drivers/MicroBitFileSystem.cpp]: Fine.  I'll stop being a purist.  We'd need a lot more flash before this becomes a real issue.

**Comment by jamesadevine** [comment]
@finneyj I was just thinking we should have tighter validation around filenames, I can currently create files with ascii characters that are incompatible with other filesystems (e.g. FAT12) 😉 

It would also semi-safeguard against bad null termination in low-level strings.

**Comment by finneyj** [comment]
:) Yes. Suggest we go with isprint() as a filter...

File: source/drivers/MicroBitFileSystem.cpp [hunk 10]
```diff
@@ -764,7 +765,7 @@ int MicroBitFileSystem::createDirectory(char const *name)
   * If a file is opened that doesn't exist, and MB_CREAT isn't passed,
   * an error is returned, otherwise the file is created.
   *
-  * @param filename name of the file to open, must be null terminated.
+  * @param filename name of the file to open, must contian only printable characters.
```
jamesadevine: Typo

source/drivers/MicroBitFile.cpp
finneyj: default open mode for MicroBitFile updated to READ | WRITE | CREATE as discussed, and an isValid() added.

source/drivers/MicroBitFileSystem.cpp
finneyj: Compatibility for ARMCC added, and tested through on mbed.org. ARMCC comes out a few hundred bytes more efficient.

inc/drivers/MicroBitFile.h
finneyj: flush() method added in both MicroBitFileSystem and MicroBitFile. 

You're quite correct - semantics are equivalent to close() followed by open().

source/drivers/MicroBitFileSystem.cpp
finneyj: fixed

finneyj:
Folks - I think this addresses all raised concerns?

source/drivers/MicroBitFile.cpp
remay: Can you give a use case for open().  As far as I can see it's not possible to create the MicrobitFile object with having opened it (which would involve creating it if necessary).   Once the object is created then the only reason for calling open() would be if I've called close() on it (and why would I do that, rather than create a new object?), and if I've called close() then the underlying file still exists, and so I don't need the CREATE flag.  The only time the underlying file doesn't exist is if I call remove() - and if I've done that adding the CREATE flag prevents me using open() as a test of existence ... I guess my real question is why the same arguments that we used for not doing this on the constructor don't apply here?

Couldn't we get the same default behaviour by doing the same as we did for the constructor by making the default for mode = READ_AND_WRITE | CREATE

Maybe I'm missing something ...

jamesadevine: @remay there is a case when you're in a loop:

(excuse the semi-pseudocode)
```
int  i = 0;
while(i < 10)
{
    MicroBitFile f("Filename" + i);
    f.write("ABC123");
    f.close();
    i++;
}
```
Naturally, some users would allocate resources outside of the loop meaning that the file instance would move from inside the loop to outside:

```
int  i = 0;
MicroBitFile f("Filename" + i);
while(i < 10)
{
    f.open("Filename" + i);
    f.write("ABC123");
    f.close();
    i++;
}
```
I think it's important to ensure that the API is flexible. It's also important to ensure that the two mechanisms are equivalent, which currently they are not. I'll push a commit in a little while to correct this.

remay: @jamesadevine But .. .open() doesn't take a filename, just a mode.  So you can't do that.  It only allows you to re-open a file that was named in the constructor.

jamesadevine:
In which case that completely makes no sense and should be removed!

On 10 Oct 2016, at 11:05, Robert May <notifications@github.com<mailto:notifications@github.com>> wrote:


@remay commented on this pull request.

In source/drivers/MicroBitFile.cpp<https://github.com/lancaster-university/microbit-dal/pull/219>:

> +  *
+  * @param mode One of: READ, WRITE, READ_AND_WRITE. Defaults to READ_AND_WRITE.
+  *
+  * @return MICROBIT_OK on success, MICROBIT_NOT_SUPPORTED if the file is already open,
+  *         MICROBIT_INVALID_PARAMETER if the filename is too large, MICROBIT_NO_RESOURCES
+  *         if the file system is full, or memory could not be allocated..
+  *
+  * @note MicroBitFiles are opened at construction and are implicitly closed at
+  *       destruction. They can be closed explicitly using the close() member function.
+  */
+int MicroBitFile::open(int mode)
+{
+    if(fileHandle >= 0)
+        return MICROBIT_NOT_SUPPORTED;
+
+    int ret = MicroBitFileSystem::defaultFileSystem->open(fileName.toCharArray(), mode | CREATE);


@jamesadevine<https://github.com/jamesadevine> But .. .open() doesn't take a filename, just a mode. So you can't do that. It only allows you to re-open a file that was named in the constructor.

finneyj:
:-) agree... can you do the honours please James?

source/drivers/MicroBitFile.cpp
jamesadevine: 3aa4543

finneyj:
OK - thanks James. That addresses all issues raised. 

Thanks to all for input here.

PR #227, opened 2016-10-21
JMX: JSON Mux

jamesadevine:
**Introducing JMX:**  an implementation of an optional serial line multiplexing protocol that allows communication between the nrf5188 and KL26 interface chip.

This PR adds JMXSerial which enables the interrogation, reading and writing of a filesystem from the USB interface chip using the serial line. It also enables the preservation of 2 flash pages that can be configured from `MicroBitConfig.h`. 

In short - it opens up a whole host of learning opportunities in the classroom.... :smile:

**Details about the protocol:**

* Uses SLIP  encoding to delimit the start and end of a packet.
* The body of a JMX packet is JSON.

**Known limitations:**

* Using specific SLIP escape characters without using the JMXSerial classes inbuilt `send` function could result in undefined behaviour. (i.e. missing serial characters)
* The JMXSerial class uses events to respond to queries, therefore if events are not serviced whilst the Interface chip is expecting a response, the request from the PC will timeout.

**Usage:**

* To share files stored on the micro:bit, enable the `MICROBIT_IF_CHIP_FS` configuration option, by setting it to 1.
* To preserve flash pages on the micro:bit, enable the `MICROBIT_IF_CHIP_PRESERVE` configuration option, by setting it to 1. The specific pages which are preserved can be configured through modifying `MICROBIT_IF_PRES_[1,2]`

[IF_BETA_21_10_16.hex.zip](https://github.com/lancaster-university/microbit-dal/files/544743/IF_BETA_21_10_16.hex.zip)


dpgeorge:
> Using specific SLIP escape characters without using the JMXSerial classes inbuilt send function could result in undefined behaviour. (i.e. missing serial characters)

How will this work with "raw" serial data that is being sent from the nrf through the kl26 to the PC?  For example, in MicroPython, if I do `uart.write(<some bytes using whole range from 0-255>)` will they get through to the PC unchanged?

Does the user need to update the kl26 firmware to use this new feature?

finneyj:
@dpgeorge 

James has sadly had a loss in his family, so I'll chip in here while he's taking the necessary personal time.

The short answer is "yes", raw serial data will get through unless you request something to the contrary.
A KL26 firmware update is needed to us this feature yes (that implements the other side of the JMX protocol). James supplied a beta version attached to the PR.

When this firmware powers up, there's a short window (a few seconds) where the firmware looks for a JMX enable handshake over the serial line from the nrf51822. If it doesn't see that (or it sees something else) it drops into the normal mbed behaviour. This therefore provides transparency for any languages/environments that don't run the JMX protocol. 

In other words, the KL26 JMX behaviour is "opt in" and transparent to any code running on the nrf518822 that doesn't do so explicitly at boot time.

dpgeorge:
> In other words, the KL26 JMX behaviour is "opt in" and transparent to any code running on the nrf518822 that doesn't do so explicitly at boot time.

Ok, thanks for the clarification.

inc/core/MicroBitConfig.h
```diff
@@ -416,6 +416,22 @@ extern uint32_t __etext;
 #define MICROBIT_HEAP_DBG                       0
 #endif
 
+// Enabling this flag will proxy File system requests on the interface chip to
+// the micro:bit, allowing easy access to files stored on the micro:bit.
+// A specific revision of the firmware is required for this to function.
```
finneyj: OK, but can we be specific about which version?

inc/core/MicroBitConfig.h
```diff
@@ -416,6 +416,22 @@ extern uint32_t __etext;
...
+#ifndef MICROBIT_IF_CHIP_FS
+#define MICROBIT_IF_CHIP_FS                     0
+#endif
+
+#define MICROBIT_IF_CHIP_VERSION                "0.0.1"
+
+// Enabling this flag will inform the interface chip to preserve MICROBIT_IF_PRES_1
+// and MICROBIT_IF_PRES_2 during the flashing of hex files.
+// A specific revision of the firmware is required for this to function.
```
finneyj: as above

inc/core/MicroBitConfig.h
```diff
@@ -436,6 +452,34 @@ extern uint32_t __etext;
...
+#endif
+
+// If we need to preserve pages, but we don't want the file system, then we
+// inform the constructor of JMXSerial to explicitly disable JMX, whilst preserving
+// the specific pages of flash memory above.
+
+// This ultimately means: pages can be preserved without the need (or overhead)
+// of file system request handlers.
+#if !CONFIG_ENABLED(MICROBIT_IF_CHIP_FS)
+#undef MICROBIT_IF_CHIP_FS
```
finneyj: Why is this pattern different to other options? why not just #ifndef MICROBIT_IF_CHIP_FS?

inc/core/MicroBitConfig.h
```diff
@@ -436,6 +452,34 @@ extern uint32_t __etext;
...
+
+#define MICROBIT_IF_CHIP_FS                     1
+#define MICROBIT_IF_CHIP_FS_OVERRIDE            1
```
finneyj: What's this for?

inc/drivers/JMXSerial.h
```diff
@@ -0,0 +1,171 @@
...
+      * Internal method to send a JSON formatted string.
+      */
+    void sendString(char* string);
+
+    /**
+      * Opens and configures the file descriptor for this JMXSerial instance.
+      * It also adds this instance to idle components, so that the fd is reset
+      * after a file has not been accessed for a short amount of time.
+      *
+      * @param fs A pointer to an instance of MicroBitFileSystem
```
finneyj: No such parameter?

inc/drivers/JMXSerial.h
```diff
@@ -0,0 +1,171 @@
...
+      * Each time the Serial module's buffer is empty, write a character if we have
+      * characters to write.
+      */
+    void dataWritten();
+
+    public:
+
+    /**
+      * Constructor.
+      * Create an instance of MicroBitSerial
```
finneyj: typo? JMXSerial?

inc/drivers/MicroBitSerial.h
```diff
@@ -56,12 +58,17 @@ enum MicroBitSerialMode
   */
 class MicroBitSerial : public RawSerial
 {
+    protected:
+    //holds the state of the baudrate for all MicroBitSerial instances.
+    static int baudrate;
```
finneyj: Why would the baud rate be the same for all serial instances?

inc/platform/Base64.h
```diff
@@ -0,0 +1,90 @@
...
+*         Return value:
+*             Returns the length of the decoded form of a
+*             base64 encoded string
+*         Requirements:
+*             1. input must not be null
+*             2. input must be greater than or equal to zero
+*/
+int base64_dec_len(char *input, int inputLen);
+
+#endif // _BASE64_H
```
finneyj: Are we actually using BASE64?

inc/platform/SLIP.h
```diff
@@ -0,0 +1,28 @@
...
+#define SLIP_ESC_ESC        '\xdd'
+
+enum SLIPCharacters {
+    END = 1,
+    ESC,
+    ESC_END,
+    ESC_ESC
+};
+
+inline int is_slip_character(char c)
```
finneyj: Do we want inline here?  Could lead to FLASH code bloat if it's used a lot?

inc/platform/jmx.h
```diff
@@ -0,0 +1,131 @@
...
+#define JSON_FIELD_MAX_LEN        8
+
+#define STATUS_OK                1
+#define STATUS_SUCCESS            0
+#define STATUS_ERROR            -1
+#define STATUS_LOCKED            -2
+#define STATUS_SLIP_ESC            -3
+
+#define STRING_INDICATOR        1
+#define NUMBER_INDICATOR        2
```
finneyj: Indentation looks funky here.

inc/platform/jmx.h
```diff
@@ -0,0 +1,131 @@
...
+int jmx_parse(char c);
+
+/**
+  * Returns the previous char, captured by jmx.
+  */
+char jmx_previous();
+
+/**
+  * Accepts a single character and updates the state of this instance of JMX parser, and also maintains a "serial_state" varaible
+  *
```
finneyj: typo - varaible.

inc/platform/jmx_actions.h
```diff
@@ -0,0 +1,216 @@
+#ifndef JMX_ACTIONS_H
+#define JMX_ACTIONS_H
+
+#include "jmx_packets.h"
+#include "jmx.h"
+#include <stdint.h>
+#include <stdlib.h>
```
finneyj: do we really need these std includes?

inc/platform/jmx_actions.h
```diff
@@ -0,0 +1,216 @@
...
+
+#define KEY_BUFFER_SIZE            10
+#define JMX_ACTION_COUNT           6
+#define JMX_TABLE_COUNT            5
+
+#define member_size(type, member)    sizeof(((type *)0)->member)
+//#define offsetof(st, m) ((size_t)&(((st *)0)->m))
```
finneyj: dead code? delete?

inc/platform/jmx_actions.h
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+  *    {
+  *        "redirect" : {
+  *            "src": "interface",
+  *            "dest": "interface"
+  *        }
+  *    }
+  */
+const JMXActionTable redirectMap = {
```
finneyj: why are some of these static const, and some not?

inc/types/ManagedBuffer.h
```diff
@@ -0,0 +1,256 @@
...
+    int readBytes(uint8_t *dst, int offset, int length, bool swapBytes = false) const;
+
+    int writeBytes(int dstOffset, uint8_t *src, int length, bool swapBytes = false);
+
+    int writeBuffer(int dstOffset, const ManagedBuffer &src, int srcOffset = 0, int length = -1);
+
+    bool isReadOnly() const { return ptr->isReadOnly(); }
+};
+
+#endif
```
finneyj: coool... we should look to see if we can lay down some other managed types (e.g. PacketBuffer)  by having this more general type.

source/drivers/JMXSerial.cpp
```diff
@@ -0,0 +1,628 @@
...
+}
+#endif
+
+/**
+  * Internal method to send a JSON formatted string.
+  */
+void JMXSerial::sendString(char* string)
+{
+    putc('"');
+    for (uint8_t i = 0; i < strlen(string); i++)
```
finneyj: uint8_t? really?

source/drivers/JMXSerial.cpp
```diff
@@ -0,0 +1,628 @@
...
+        putc(string[i]);
+    putc('"');
+}
+
+/**
+  * Opens and configures the file descriptor for this JMXSerial instance.
+  * It also adds this instance to idle components, so that the fd is reset
+  * after a file has not been accessed for a short amount of time.
+  *
+  * @param fs A pointer to an instance of MicroBitFileSystem
```
finneyj: again, doesn't exist...

source/drivers/JMXSerial.cpp
```diff
@@ -0,0 +1,628 @@
...
+    if(fd < 0)
+    {
+        fd = MicroBitFileSystem::defaultFileSystem->open(file, mode);
+
+        if(fd < 0)
+            return MICROBIT_INVALID_PARAMETER;
+
+        filename = file;
+
+        if(status == 0)
```
finneyj: should this check a flag?

source/drivers/JMXSerial.cpp
```diff
@@ -0,0 +1,628 @@
...
+
+        if(p->len - i > EFFECTIVE_BUFF_SIZE)
+        {
+            putc(SLIP_ESC);
+            putc('A');
+        }
+
+        MicroBitFileSystem::defaultFileSystem->write(fd, buf, rx_size);
+
+        i += rx_size;//
```
finneyj: rogue comment?

source/drivers/MicroBitFileSystem.cpp
```diff
@@ -211,7 +211,7 @@ int MicroBitFileSystem::load()
         return MICROBIT_NO_DATA;
 
     rootDirectory = root;
-    fileSystemSize = root->length;
+    fileSystemSize = root->length / MBFS_BLOCK_SIZE;
```
finneyj: Is this a bugfix?

source/platform/Base64.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+inline unsigned char b64_lookup(char c) {
+    if (c >= 'A' && c <= 'Z') return c - 'A';
+    if (c >= 'a' && c <= 'z') return c - 71;
+    if (c >= '0' && c <= '9') return c + 4;
+    if (c == '+') return 62;
+    if (c == '/') return 63;
+    return -1;
+}
```
finneyj: Again, do we actually use BASE64 encoding?

source/platform/jmx.cpp
```diff
@@ -0,0 +1,653 @@
...
+#include "mbed.h"
+#include "MicroBitHeapAllocator.h"
+#include "MicroBitCompat.h"
+
+JMXInitPacket* jmx_init_p = NULL;
+FSRequestPacket* jmx_fsr_p = NULL;
+DIRRequestPacket* jmx_dir_p = NULL;
+StatusPacket* jmx_status_p = NULL;
+RedirectPacket* jmx_redirect_p = NULL;
+UARTConfigPacket* jmx_uart_p = NULL;
```
finneyj: can these be made static?

p.s. what's the rationale for this bit being purely C rather than cpp? Nothing wrong with that, just less consistent with the restof the code.

source/platform/jmx.cpp
```diff
@@ -0,0 +1,653 @@
...
+extern void user_putc(char c);
+
+extern void jmx_packet_received(char*);
+
+void nop(void*) {};
+#ifdef __cplusplus
+}
+#endif
+
+__inline int jmx_get_action_index(JMXActionTable* t, char* key, int key_len)
```
finneyj: These are perhaps quite hefty for inline functions?

PR #228, opened 2016-10-21
Update dependencies for older gcc compilation for MicroPython.

carlosperate:
Basically we need to update the nRF51822 version to `nrf51822#v2.5.0+mb6` containing this commit https://github.com/lancaster-university/nRF51822/commit/c5ce428daede8d939c4496453e84d03955c99804 to resolve https://github.com/bbcmicrobit/micropython/issues/338 (more info on https://github.com/ARMmbed/ble-nrf51822/issues/133).

This PR is targetting `www-microbit-co-uk`, but ideally I'd suggest to create a new branch from tag `v1.4.19` and and then change this PR base branch to it (GitHub has added that [feature](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) not that long ago). Or do this manually, whichever you prefer.

### DIFF
--- a/module.json
+++ b/module.json
@@ -1,6 +1,6 @@
 {
   "name": "microbit-dal",
-  "version": "1.4.19",
+  "version": "1.4.20",
   "license": "MIT",
   "description": "The runtime library for the BBC micro:bit, developed by Lancaster University",
   "keywords": [
@@ -16,8 +16,8 @@
   "dependencies": {
     "mbed-classic": "lancaster-university/mbed-classic#microbit_hfclk",
     "ble": "lancaster-university/BLE_API#v2.5.0+mb3",
-    "ble-nrf51822": "lancaster-university/nrf51822#v2.5.0+mb5",
-    "nrf51-sdk": "lancaster-university/nrf51-sdk#v2.2.0+mb3"
+    "ble-nrf51822": "lancaster-university/nrf51822#v2.5.0+mb6",
+    "nrf51-sdk": "lancaster-university/nrf51-sdk#v2.2.0+mb4"
   },
   "extraIncludes": [
     "inc"


dpgeorge:
@finneyj this would fix a long-standing (~few months old) problem that people have compiling our code.

finneyj:
@dpgeorge @carlosperate 

Yes, of course. 1.4.20 has already been tagged to support a request from BBC to add accessibility support though (and contains some init code to support this that I suspect you don't want in micropython at the moment). A branch at 1.4.19 sounds like the best option for a small change.

We stopped active development on 1.4.x in February, so it's just you and the BBC editors that are on there now. So if we fork, that branch would be all yours. Any preferred name for the branch? It will show up in your yotta configuration, so I thought you might like a say in that. :-)

dpgeorge:
Thanks!  I suggest a tag of 1.4.19b (it's a very minimal update to 1.4.19, and it makes sense that it has a name that comes "before" the pre-existing 1.4.20).

finneyj:
no problem. You could have a micropython branch name if you want though, if you think you might need further tweaks / configurations etc.

finneyj:
@carlosperate @dpgeorge 

I've created a micropython branch alongside www-microbit-co-uk, as it's useful to ease maintenance, and tagged v1.4.19b as requested. It builds ok as a standalone module, but I haven't tested against a micropython firmware build.

Let me know when if it's ok to close this issue.

carlosperate:
Thanks Joe, I could test it tonight just to make sure, and confirm here.

finneyj:
perfect - thanks @carlosperate.

carlosperate:
I can confirm v1.4.19b works as expected :)

finneyj:
no problem - thanks for confirming Carlos. :)

PR #230, opened 2016-10-25
Guarded against redefinition of MICROBIT_BLE_SECURITY_LEVEL which cou…

bluetooth-mdw:
…ld occur if the bluetooth open config.json flag was set


finneyj:
Thanks @bluetooth-mdw

PR #247, opened 2016-12-03
Quadrature decoder hardware driver.

sh1boot:
I'm only moving things around in circles, now, so I guess I should put it up for review.

Next mission will be a [software implementation](https://github.com/sh1boot/testqdec/blob/master/source/SoftQDec.cpp) of the same thing (using pin edge events to detect motion), because one decoder isn't sufficient keep a two-wheeled buggy going in a straight line.  That's what the `virtual`s are about.

### DIFF
--- a/inc/drivers/MicroBitQuadratureDecoder.h
+++ b/inc/drivers/MicroBitQuadratureDecoder.h
@@ -1,0 +1,174 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016-2017 Simon Hosie
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef MICROBIT_QUADRATUREDECODER_H
+#define MICROBIT_QUADRATUREDECODER_H
+
+#include "mbed.h"
+#include "MicroBitConfig.h"
+#include "MicroBitComponent.h"
+#include "MicroBitPin.h"
+
+                                                        // Configuration flags...
+#define QDEC_USING_SYSTEM_TICK              0x01        // Use systemTick() to keep position up to date.
+#define QDEC_USING_DEBOUNCE                 0x02        // Use input debounce feature.
+#define QDEC_LED_ACTIVE_LOW                 0x04        // Drive LED pin low to activate.
+
+/**
+  * Class definition for MicroBit Quadrature decoder.
+  *
+  */
+class MicroBitQuadratureDecoder : public MicroBitComponent
+{
+    protected:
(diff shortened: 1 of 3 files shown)


sh1boot:
Next mission will be a software implementation of the same thing (using pin edge events to detect motion), because one decoder isn't sufficient keep a two-wheeled buggy going in a straight line  That's what the `virtual`s are about.

inc/drivers/MicroBitQuadratureDecoder.h
```diff
@@ -0,0 +1,175 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 British Broadcasting Corporation.
+This software is provided by Lancaster University by arrangement with the BBC.
```
sh1boot: Oops copy-paste.

I don't really want to own copyright on anything; what's the proper default copyright notice to put on new files?

source/drivers/MicroBitQuadratureDecoder.cpp
```diff
@@ -0,0 +1,234 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 British Broadcasting Corporation.
+This software is provided by Lancaster University by arrangement with the BBC.
```
sh1boot: Copy-pasted copyright again...

source/drivers/MicroBitQuadratureDecoder.cpp
```diff
@@ -0,0 +1,233 @@
...
+    __NOP();
+    __NOP();
+
+    NRF_QDEC->TASKS_READCLRACC = 1; // Clear accumulators
+    NRF_QDEC->ENABLE = 1;
+
+    NRF_QDEC->TASKS_START = 1;
+    status |= MICROBIT_COMPONENT_RUNNING;
+
+    if ((flags & QDEC_USING_SYSTEM_TICK) != 0)
```
jamesadevine: Don't you want to enable the system tick here?

source/drivers/MicroBitQuadratureDecoder.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    return MICROBIT_OK;
+}
+
+/**
+  * Stop the hardware and make it available for use by other instances.
+  */
+void MicroBitQuadratureDecoder::stop()
+{
+    if ((flags & QDEC_USING_SYSTEM_TICK) != 0)
```
jamesadevine: call disableSystemTick? You don't unset the flag here.

inc/drivers/MicroBitQuadratureDecoder.h
```diff
@@ -0,0 +1,174 @@
...
+#define QDEC_LED_ACTIVE_LOW                 0x04        // Drive LED pin low to activate.
+
+/**
+  * Class definition for MicroBit Quadrature decoder.
+  *
+  */
+class MicroBitQuadratureDecoder : public MicroBitComponent
+{
+    protected:
+    static MicroBitPin pinNC;           // A no-connect pin -- the default for unused LED arguments.
```
sh1boot: I was thinking then that multiple driver instances would then imply multiple `pinNC` instances; and that the linker would throw it away if it wasn't used.  However, leaving it to the linker does mean that the object lives forever even if this driver is only used sometimes (if it's never used then the object should vanish).

How big is it?  Is it possible to put it in ROM?  In ROM it's only as bad as all the other code that is present because it'll sometimes be used.

jamesadevine: Yeah, I can see you considered optimisation. Realistically how many of these instances will you run at one time? I think constantly holding it in either RAM or Flash seems a bit unnecessary if you're likely to have < 2 instances.

source/drivers/MicroBitQuadratureDecoder.cpp
sh1boot: Oops, yes I do!

source/drivers/MicroBitQuadratureDecoder.cpp
sh1boot: I think I've named the flag badly.  It should be `QDEC_USE_SYSTEM_TICK`.  It's a configuration item that comes in through the constructor or through `disable`/`enableSystemTick()`.  `stop()` should just obey the flag and unregister the hook if it's currently registered.

Perhaps it's more appropriate to keep a flag in `status` confirming when the hook is registered, rather than registering and unregistering it based on what the configuration implies that we should have done in the past.  Though I'm not sure if the extra complexity offers any real value.

inc/drivers/MicroBitQuadratureDecoder.h
sh1boot: I do wonder, if the future involves many drivers for plug-in peripherals, could other drivers also need a generic "no-connect" referent available to their constructors?

I think for now I should make `LED` a pointer, not a reference, and initialise it to `NULL` rather than `&pinNC`.  As a reference it's a bit fiddly to initialise to dynamic storage, and I'd probably end up having another pointer to maintain besides.

jamesadevine:
@sh1boot looks good to me :smile: great job!

Issue #252, opened 2016-12-19
bug in MicroBitUARTService#read when buffer contains NULL character


kartben:
When reading the required number of bytes, MicroBitUARTService#read() will stop on a null character - I believe this is wrong as nothing in the doc says that this method should treat the buffer as a null-terminated string.
If people want to read() strings, they can use the signature that returns a ManagedString.

kartben:
PR #253 has been merged! 
Closing.

PR #253, opened 2016-12-19
fix for issue #252: bug in MicroBitUARTService#read when buffer conta…

kartben:
…ins NULL character

Signed-off-by: Benjamin Cabé <benjamin@eclipse.org>

### DIFF
--- a/source/bluetooth/MicroBitUARTService.cpp
+++ b/source/bluetooth/MicroBitUARTService.cpp
@@ -356,7 +356,7 @@ int MicroBitUARTService::read(uint8_t *buf, int len, MicroBitSerialMode mode)
     {
         int c;
 
-        while((c = getc(mode)) > 0 && i < len)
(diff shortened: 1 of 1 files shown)


martinwork:
Should the && operands be swapped as well, to avoid reading and discarding an extra character after len have been read?

jamesadevine:
Yup, @martinwork . We really need a Stream abstraction, it's been on our todo list for a while. This is another place where the same broken logic has cropped up again, I suspect because it's been implemented across 3 different classes, and the first time it was caught the change wasn't applied here.

kartben:
Good catch @martinwork. I updated the condition.

kartben:
any plan to merge this?

finneyj:
Apologies @kartben - my many hands are full as usual. :-)

Looks great, very happy to merge. Thanks for the patch.

PR #290, opened 2017-05-16
Sound extensions

utz82:
Added a new class for creating sample recordings from microphone input, and added sample playback capabilities to CalliopeSoundMotor.

The sample format is optimized for 1-bit output, and allows for efficient data compression at the cost of sound quality. Efficiency depends on the harmonic spectrum of the sample, with higher frequencies generating larger data. On average, 2-3 seconds of human speech can be compressed into 1 KB of sample data at the default sample rate. Sample format is unsigned 8-bit PWM, which is a simple record of the time taken until the next output state change. If a sample byte is 0, then no state change occurs and the next state change is delayed by 256 ticks. Samples are recorded and played back at a rate of 8 KHz by default, this can be changed to up to ~11KHz. At faster rates, correct playback is not ensured.

Both recording and playback are implemented as interrupt services. CPU load will be approx. 85% at 8 KHz, so it is not possible to run other load-heavy tasks in parallel.

Sampling thresholds for recording are tuned to the current board setup, they will need to be adjusted once the planned changes to the board are implemented. Likewise, there will be some minor changes necessary if the piezo is replaced by a true speaker.

This is my first contribution to the Calliope mini project, please be patient with me and let me know about any mistakes. Also note that I have not changed the versioning.


finneyj:
No problem @utz82 - this looks like really interesting work! 

Please do let us know how you get on with your project, as I imagine lots of people from the micro:bit and calliope communities would be interested in your results.

utz82:
Admittedly most of those changes just reflect the state of the current Calliope fork of microbit-dal. The only contribution from my side is some sampled sound support for the Calliope's microphone/speaker set-up, which is probably of little use for the original microbit. However, I also have plans to add polyphonic sound synthesis later on. It should be possible to adapt this for the microbit (with a speaker connected to pin0/gnd).

PR #292, opened 2017-05-25
Bugfix 289

### LINKED ISSUE #289: MicroBitIOPinService truncates analogue input values above 255
onDataRead and idleTick both do this: 

uint8_t value;
value = io.pin[i].getAnalogValue(); 

So values above 255 get reported mod 256.

finneyj:
Fix #289

### DIFF
--- a/inc/bluetooth/MicroBitIOPinService.h
+++ b/inc/bluetooth/MicroBitIOPinService.h
@@ -122,15 +122,7 @@ class MicroBitIOPinService : public MicroBitComponent
       * @param i the enumeration of the pin to test
       * @return 1 if this pin is configured as an input, 0 otherwise
       */
-    int isInput(int i);
-
-    /**
-      * Determines if the given pin was configured as output by the BLE IOPinConfigurationCharacterisitic.
-      *
-      * @param i the enumeration of the pin to test
-      * @return 1 if this pin is configured as an output, 0 otherwise
-      */
-    int isOutput(int i);
+    int isActiveInput(int i);
 
     /**
      * Scans through all pins that our BLE client have registered an interest in. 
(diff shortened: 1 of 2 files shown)


cpseager:
LGTM. Whilst working on this is it possible to rename the really confusing isInput and isOutput as well?

These are actually nothing to do with a DDR as both the name and comments suggests - this really confused me when first using them. Setting/clearing the bit does **not** configure the pin as an input or as an output! Setting the pin simply 'activates' that pin as an input that is read rather than ignored..

isInput should be something like isInputActive

isOutput should be something like isInputInactive

finneyj:
@cpseager  sure - done.

finneyj:
@martinwork Anything you'd like to add?

cpseager:
@finneyj Thanks. Update line 49 comment too?

martinwork:
Thanks @finneyj . Creating updateBLEInputs was a good idea. However this made me notice that onDataRead is calling notify. I'm far from sure and I am not in a position to easily test this right now, but I suspect calling notify is not appropriate inside the ReadAuthorization callback, because a read is already underway.

How quickly is this fix likely to make it to PXT?

finneyj:
@martinwork Thanks. 

I think this is ok.  Recall that the notify() call doesn't necessarily raise a notification, but records the new value of the characteristic flags with the BLE stack that a notification should be scheduled if appropriate.  It worked in my tests anyway (I tested both reads and notify via the Nordic diagnostic tools).

Once we merge it'll go into the next release and pretty much straight into PXT (usually a few days lag to beta, then trickle in live). I want to merge in the compass and FS bug fixes to next release too.

finneyj:
Thanks all - updated and merged

Issue #286, opened 2017-04-22
File system causing corruptions


DavidWhaleMEF:
microbit.org-ticket: 1229

The files can be read out many times BUT a Reset or Power Down will clear the data.

I seem to remember reading that the data file was kept in EEPROM and NOT destroyed by a Reset or Power Down. Can someone please confirm that this statement is true?

DavidWhaleMEF:
See also: https://github.com/Microsoft/pxt/issues/1495

finneyj:
@DavidWhaleIET 

Correct - all data is held in FLASH memory so is persistent. curious.

Let me try to repro your code in C.

finneyj:
Just tested this on native C, which is a pretty close replica of how PXT uses the FileSystem classes, and it seems to work ok. Certainly persists data across a reset and no data corruption that I can see.

Have you got a copy of the PXT code so I can repro?

```
#include "MicroBit.h"
#include "MicroBitFileSystem.h"
#include "MicroBitFile.h"


MicroBit uBit;
MicroBitImage heart("0,255,0,255,0\n255,255,255,255,255\n255,255,255,255,255\n0,255,255,255,0\n0,0,255,0,0\n");

const char *fname = "output.txt";

void initFileSystem()
{
    if (MicroBitFileSystem::defaultFileSystem == NULL)
        new MicroBitFileSystem();
}

void appendLine(const char* filename, const char* text)
{
    initFileSystem();
    ManagedString fn(filename);
    ManagedString t(text);
    MicroBitFile f(fn);
    f.append(t);
    f.append("\r\n");
    f.close();
}

void readToSerial(const char* filename)
{
    initFileSystem();
    ManagedString fn(filename);
    MicroBitFile f(fn);

    while(1)
    {
        int c = f.read();
        if (c <= 0)
            break;

        uBit.serial.printf("%c", (char)c);
    }

    f.close();
}

void onButtonA(MicroBitEvent)
{
    appendLine(fname, "hello");
    uBit.serial.printf("W");
    uBit.display.print("W");
}


void onButtonB(MicroBitEvent)
{
    readToSerial(fname);
    uBit.display.print("H");
}

int main()
{
    // Initialise the micro:bit runtime.
    uBit.init();

    uBit.messageBus.listen(MICROBIT_ID_BUTTON_A, MICROBIT_BUTTON_EVT_CLICK, onButtonA);
[...]

PR #293, opened 2017-05-27
Erase scratch pages after use [#286]

finneyj:
MicrobitFileSystem occasionally needs to "refresh" FLASH page that have been
marked as deleted, such that they can be used again. When this happens, an
unused FLASH page is used to temporarily hold wanted data (a scratch page).
This is alloctaed ina round robin fashion from all available data pages to
load balance wear across pages.

This patch ensures that a page used in this fashion is erased after use, such
that it is ready for use again as a data page.

This patch may also address https://github.com/Microsoft/pxt/issues/1495


DavidWhaleMEF:
Can you let me know when this surfaces in pxt/beta so we can test it, thanks.

pelikhan:
It is in microbit 0.12.22

DavidWhaleMEF:
Great, thanks :)

Issue #269, opened 2017-02-12
MIT licence missing from some files


whaleygeek:
The MIT licence header is missing from these files:

drivers/MicroBitFile.h 
drivers/MicroBitFileSystem.h
drivers/MicroBitFlash.h

drivers/MicroBitFile.cpp
drivers/MicroBitFileSystem.cpp
drivers/MicroBitFlash.cpp

finneyj:
Thanks. Will fix.

alennie1993:
Added in and PR sent, closing issue.

Issue #287, opened 2017-04-23
Minor spelling mistake


DavidWhaleMEF:
Just noticed this, while looking something up

https://github.com/lancaster-university/microbit-dal/blob/master/source/core/MicroBitDevice.cpp#L80

Says 'Derived' should say 'Derive' or 'Get a derived' ??

alennie1993:
Committed and PR sent, closing issue.

PR #298, opened 2017-06-13
Quick fixes and touches

alennie1993:
Just Adding MIT Licences to some files that were missing as per Issue #269


alennie1993:
#287 , Quick spelling fix.

finneyj:
Thanks for the PR @alennie1993!

Looks good to me, just one minor thing. Can you update commit messages to be a bit more specific? for "MicroBitFlash.h"? that one will look weird in the history as is...

PR #303, opened 2017-07-23
added support for Write Without Response to Client Event characteristic

smartyw:
Write Without Response gives much higher throughput than Write requests, but with no ACKs. Much better for some types of application.


finneyj:
Thanks @smartyw. 

Looks like this PR let's the BLE client decide if events it has generated should be acknowledged or not, rather than enforcing acknowledgements as a policy. Good decision i think. :-)

smartyw:
Yes, that's correct. The event service is versatile and useful in many situations. Some warrant communication with ACKs (Write) whilst for others, throughput is more important and that's when Write Without Response should be used. 

Nice thorough answer for the benefit of others :-)

smartyw:
@finneyj  @pelikhan 

Is this in MakeCode yet?

pelikhan:
No. Is there a release tag?

bluetooth-mdw:
Thanks @pelikhan 

I guess not then! Not clear on how the release process works to be honest. I can pick this up with @finneyj  when he's next around.

smartyw:
@jaustin I'm not clear how the release process works for getting changes into MakeCode. Do you need anything else for this change to be released?

smartyw:
@finneyj @pelikhan This still seems to not have made it into MakeCode. Are you waiting for me or @bluetooth-mdw to do something? 

@jaustin could we please document the release procedure so that contributors know what they're meant to do to ensure PRs get released across the various tools? It seems sub-optimal and confusing for people to have code created by one tool potentially behave differently to the equivalent created by another tool.

PR #312, opened 2017-08-16
Reduce default advertising interval and make property of config.json

smartyw:
I think the advertising interval of 200ms is too slow and contributes to some old (and slow) smartphones from struggling to find micro:bits when scanning. I just tested and the difference between advertising at the current 200ms interval and the minimum of 20ms was startling with my old Moto G 2nd gen. At 200ms the Moto G frequently failed to find the micro:bit after 5 seconds of scanning. At 20ms it found it near-instantaneously every time.

Exposing this as a property gives us flexibility. I propose that the default be reduced to say 50ms though.


jaustin:
Thanks for this suggestion, I think we should do the default through the MicroBitConfig.h file so that people using !yotta can also specify this (or people overriding config elsewise)

@finneyj is this a good task for someone over the summer?

smartyw:
It should be in both MicroBitConfig.h and config.json. It was an oversight not to include it in MicroBitConfig.h. I can add and resubmit the PR if you like....?

smartyw:
@jaustin added to MicroBitConfig.h

Hope to see this merged and released to MakeCode soon. This made a big difference in testing with old and slow phones.

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -414,8 +414,11 @@ void MicroBitBLEManager::init(ManagedString deviceName, ManagedString serialNumb
 
     ble->accumulateAdvertisingPayload(GapAdvertisingData::COMPLETE_LOCAL_NAME, (uint8_t *)BLEName.toCharArray(), BLEName.length());
     ble->setAdvertisingType(GapAdvertisingParams::ADV_CONNECTABLE_UNDIRECTED);
-    ble->setAdvertisingInterval(200);
-
+    uint16_t adv_interval = 50;
```
jaustin: I think now that you've put the config in MicroBitConfig.h so we're always guaranteed to have the symbol defined and it will default to 50, this can just become
```C++
ble->setAdvertisingInterval(MICROBIT_BLE_ADVERTISING_INTERVAL);
```

smartyw:
@jaustin refactored per your request. Note though that the comment you made applies to other config properties used in the MicroBitBLEManager class, so I was just being consistent. No time to go through all the existing properties right now though.

Cheers

jaustin:
Ahh, thanks.

As for the fact the pattern is used elsewhere - Hmm, there may well be a reason that @finneyj has it the way he has, so let's wait for an ack from him and then we can call it.

finneyj:
Thanks @smartyw @jaustin. Been away on a long overdue vacation so apologies this was hanging for a while.

All looks good to me. 

- Code optimization to remove that variable is fine - I imagine that pattern was there for a nordic SDK call at some point as many of them pass params by pointers. I like the simplification.
- Adding a config option for this does indeed add flexibility at no real cost so is also a win.
- As we have evidence that reducing advertising interval improves compatibility, changing the default interval from 200ms to 50ms is also a good idea IMHO. It seems unlikely this would cause any significant issues, and the impact on battery consumption will not be great.

Happy to merger this one. Thanks both.

Issue #288, opened 2017-05-02
Compass results are inaccurate in some circumstances


jaustin:
This is an issue to track some of the increased user-support requests we've had since Dr Who Live Lesson raised the priority of the compass.

Issue as users experience it: 
The compass doesn't give a full range of values, and appears to be 'stuck' in one direction

Our current understanding:
There are two confounding factors:
1) The micro:bit compass is very affected by the battery and other ferrous metals in the environment. Calibration should (must?) be performed in the same situation as the compass will be used. For example, people very rarely see issues when calibrating over USB and using on USB.

2) In certain circumstances the calibration algorithm chooses an incorrect (read: miles away) centre for the location of the calibration test points. This is *most* common when the battery is attached because the battery transforms the usual calibration point-cloud from 'points on the surface of a sphere' to 'points on the surface of a brioche bun shape' (Wikipedia tells me this is an oblate spheroid, but brioche bun will do). This increases the liklihood that the calibration algorithm will find an incorrect centre.

@finneyj has some good 3D plots, to add. Here are some images from @DavidWhaleMEF that show the issue:

Calibrated on USB, used on USB:
![redo_cal_usb_use_usb](https://cloud.githubusercontent.com/assets/147113/25615058/01141ac2-2f2d-11e7-83b8-f39b7249231e.png)

Calibrated on USB, used with battery:
![redo_cal_usb_use_bat](https://cloud.githubusercontent.com/assets/147113/25615069/17ec49fe-2f2d-11e7-89fd-e69309d66d13.png)

Calibrated on battery, used on battery

![redo_cal_bat_use_bat](https://cloud.githubusercontent.com/assets/147113/25615079/22f5ce92-2f2d-11e7-8569-ec113445f609.png)

The current assumption is that we can remedy this with some adjustments to the calibration algorithm to detect a pathological calibration failure.

jaustin:
@markshannon @dpgeorge headsup

finneyj:
@jaustin Thanks for raising here. Our foundation helpdesk tickets can be a bit hidden sometimes.

I spent a day or two analysing this and found a number of factors that contribute to this issue and generally impact on performance. Documenting results here...

1) The current least-mean-squares algorithm sometimes gets it really wrong.

[...]

Issue #318, opened 2017-09-28
"Pairing mode" could become a more generic 'utility mode'


jaustin:
(this issues should theoretically be filed against microbit not microbit-dal but I think this is a better location for eyes/discussion - notice that this would still likely be *implemented* in the microbit class so that people integrating the DAL would be able to optionally take the behaviour)

The current 'pairing mode' is already mis-named - it's used for BOTH pairing and also updating the hex file when the micro:bit is running a program with the BLE radio.

However there are other things people might like to do that are 'under the hood' to their program
* Measure and display the battery charge level (https://os.mbed.com/users/MarceloSalazar/notebook/measuring-battery-voltage-with-nordic-nrf51x/ )
* Send files from the filesystem over USB serial
* Erase compass calibration

More dramatically, if flash allowed we could even include the BLE services and aserial messagebus bridge so that a more complete 'remote control mode' could be achieved without having to re-install a program.

@finneyj and @jamesadevine I know this has long been in your minds, and @martinwork and @bluetooth-mdw the idea of clarifying that 'PAIRING MODE' is *not* only about pairing is important.

jaustin:
Note for battery level, the reading is quite simple - here's an example: https://os.mbed.com/teams/microbit/code/microbit-battery-test/

drmpf:
Can we change the 'standard' microbit ADC to use this band gap ref as well.  Currently the ADC output drops as the battery volts drop.

jaustin:
@drmpf changing what happens by default is an issue because it'd be a breaking change for a bunch of accessories, etc.

Consider the fairly common case of a slide-switch potentiometer wired off the edge connector - using ADC readings relative to VDD is more useful in that situation...

I think if we were to do the following
* Expose the ability to set the ADC reference voltage
* Expose the ability to set the ADC's input to the supply as well as the pins

then we'd be able to let people write more useful scripts. @dpgeorge may have a different ADC implementation in MicroPython, and I think we should do similar there. 

I've spawned #322 to track the ADC flexibility side and we can continue the 'utility mode' discussion here.

jaustin:
Also, when the micro:bit battery is low, flashing over BLE can be inconsistent or pairing won't work. We could set a battery threshold below which we abort pairing mode and instead say "battery too low"?

martinwork:
Low battery level can cause BLE flashing to fail halfway, leaving micro:bit soft bricked. It would be good to have the ability for the client app to check battery level before flashing.

PR #319, opened 2017-10-03
Fix compass calibration and improve UX for calibration 'game'

jaustin:
This PR brings in Joe's fixed compass calibration routine developed in #288 and adds the 'tilt to fill screen' compass calibration routine.

There will be a number of knock-on effects of changing this for documentation, so @microbit-mark and @whaleygeek here's a heads up - we can merge to master before doing this but I think the change shouldn't hit microbit live editors until we've made updates.

- [x] Update micro:bit knowledge base: https://support.microbit.org/support/solutions/articles/19000008874-calibrating-the-micro-bit-compass-what-does-it-mean-when-the-micro-bit-says-draw-a-circle-or-tilt
- [x] Update MakeCode page (https://makecode.microbit.org/reference/input/calibrate-compass )
- [ ] Liaise with @dpgeorge about the right approach to integrating this with MicroPython so we make the change at about the same time (still considering a separate patch to break out the calibration from the UX)
- [ ] Wider testing

### DIFF
--- a/source/drivers/MicroBitCompassCalibrator.cpp
+++ b/source/drivers/MicroBitCompassCalibrator.cpp
@@ -48,7 +48,7 @@ DEALINGS IN THE SOFTWARE.
 MicroBitCompassCalibrator::MicroBitCompassCalibrator(MicroBitCompass& _compass, MicroBitAccelerometer& _accelerometer, MicroBitDisplay& _display) : compass(_compass), accelerometer(_accelerometer), display(_display)
 {
     if (EventModel::defaultEventBus)
-        EventModel::defaultEventBus->listen(MICROBIT_ID_COMPASS, MICROBIT_COMPASS_EVT_CALIBRATE, this, &MicroBitCompassCalibrator::calibrate, MESSAGE_BUS_LISTENER_IMMEDIATE);
+        EventModel::defaultEventBus->listen(MICROBIT_ID_COMPASS, MICROBIT_COMPASS_EVT_CALIBRATE, this, &MicroBitCompassCalibrator::calibrateUX, MESSAGE_BUS_LISTENER_IMMEDIATE);
 }
 
 /**
@@ -80,6 +80,25 @@ int MicroBitCompassCalibrator::measureScore(CompassSample &c, CompassSample *dat
     return (maxD - minD);
 }
 
+/**
+ * Calculates an independent X, Y, Z scale factor and centre for a given set of data points, assumed to be on
+ * a bounding sphere
+ *
+ * @param data An array of all data points
+ * @param samples The number of samples in the 'data' array.
+ *
+ * This algorithm should be called with no fewer than 12 points, but testing has indicated >21 points provides
+ * a more robust calculation.
+ *
+ * @return A calibration structure containing the a calculated centre point, the radius of the minimum
+ * enclosing spere of points and a scaling factor for each axis that places those points as close as possible
+ * to the surface of the containing sphere.
+ */
+CompassCalibration MicroBitCompassCalibrator::calibrate(CompassSample *data, int samples)
+{
+    CompassSample centre = approximateCentre(data, samples);
+    return spherify(centre, data, samples);
+}
 /**
  * Calculates an independent scale factor for X,Y and Z axes that places the given data points on a bounding sphere
  *
@@ -233,7 +252,7 @@ CompassSample MicroBitCompassCalibrator::approximateCentre(CompassSample *data, 
  *
  * This function is, by design, synchronous and only returns once calibration is complete.
  */
-void MicroBitCompassCalibrator::calibrate(MicroBitEvent)
+void MicroBitCompassCalibrator::calibrateUX(MicroBitEvent)
 {
     struct Point
     {
@@ -335,9 +354,7 @@ void MicroBitCompassCalibrator::calibrate(MicroBitEvent)
         wait_ms(100);
     }
 
-    CompassSample centre = approximateCentre(data, samples);
-    CompassCalibration cal = spherify(centre, data, samples);
-    compass.setCalibration(cal);
+    compass.setCalibration(calibrate(data, samples));
 
     // Show a smiley to indicate that we're done, and continue on with the user program.
     display.clear();
--- a/source/drivers/MicroBitCompassCalibrator.cpp
+++ b/source/drivers/MicroBitCompassCalibrator.cpp
@@ -258,7 +258,6 @@ void MicroBitCompassCalibrator::calibrateUX(MicroBitEvent)
     {
         uint8_t x;
         uint8_t y;
-        uint8_t on;
     };
 
     const int PERIMETER_POINTS = 25;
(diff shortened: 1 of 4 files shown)


source/drivers/MicroBitCompassCalibrator.cpp
```diff
@@ -67,25 +242,27 @@ void MicroBitCompassCalibrator::calibrate(MicroBitEvent)
...
+
     const int PIXEL1_THRESHOLD = 200;
-    const int PIXEL2_THRESHOLD = 800;
+    const int PIXEL2_THRESHOLD = 680;
 
     wait_ms(100);
 
-	Matrix4 X(PERIMETER_POINTS, 4);
-    Point perimeter[PERIMETER_POINTS] = {{1,0,0}, {2,0,0}, {3,0,0}, {4,1,0}, {4,2,0}, {4,3,0}, {3,4,0}, {2,4,0}, {1,4,0}, {0,3,0}, {0,2,0}, {0,1,0}};
+    Point perimeter[PERIMETER_POINTS] = {{0,0,0}, {1,0,0}, {2,0,0}, {3,0,0}, {4,0,0}, {0,1,0}, {1,1,0}, {2,1,0}, {3,1,0}, {4,1,0}, {0,2,0}, {1,2,0}, {2,2,0}, {3,2,0}, {4,2,0}, {0,3,0}, {1,3,0}, {2,3,0}, {3,3,0}, {4,3,0}, {0,4,0}, {1,4,0}, {2,4,0}, {3,4,0}, {4,4,0}};
```
dpgeorge: Should these be static const so they don't take up RAM/stack?

source/drivers/MicroBitCompassCalibrator.cpp
```diff
@@ -67,25 +242,27 @@ void MicroBitCompassCalibrator::calibrate(MicroBitEvent)
...
     MicroBitImage img(5,5);
     MicroBitImage smiley("0,255,0,255,0\n0,255,0,255,0\n0,0,0,0,0\n255,0,0,0,255\n0,255,255,255,0\n");
+
+    CompassSample data[PERIMETER_POINTS];
     int samples = 0;
 
     // Firstly, we need to take over the display. Ensure all active animations are paused.
     display.stopAnimation();
-    display.scrollAsync("DRAW A CIRCLE");
+    display.scrollAsync("TILT TO FILL SCREEN");
```
dpgeorge: This new message is longer than before and so takes more time to scroll.  It can be very annoying especially when you've read it before!  I suggest 1) make it shorter, like "TILT TO FILL"; 2) allow the user to press button A or B to stop the message and go straight into the calibration game.

pelikhan: Agreed. The slowness of calibration is the number 1 reason to avoid this feature. I suggest a distincitve little animation instead of text. Maybe some kind of radar-like anim.

DavidWhaleMEF: Text also doesn't work well globally with lots of language variants. Icons and animations are easier to communicate without strong language skills via a simple video or animation on our website.

DavidWhaleMEF: How about an arrow that spins around the screen to signify a compass, followed by a question mark? Play the animation once, then the first movement the device sees after that movement triggers the actual calibration.

Something like this...
```
let moved = false
input.onGesture(Gesture.Shake, () => {
    moved = true
})
basic.forever(() => {
    moved = false
    while (!(moved)) {
        images.arrowImage(ArrowNames.North).showImage(0)
        images.arrowImage(ArrowNames.NorthEast).showImage(0)
        images.arrowImage(ArrowNames.East).showImage(0)
        images.arrowImage(ArrowNames.SouthEast).showImage(0)
        images.arrowImage(ArrowNames.South).showImage(0)
        images.arrowImage(ArrowNames.SouthWest).showImage(0)
        images.arrowImage(ArrowNames.West).showImage(0)
        images.arrowImage(ArrowNames.NorthWest).showImage(0)
        basic.showString("?")
        basic.pause(250)
        basic.clearScreen()
        basic.pause(250)
        basic.showString("?")
        basic.pause(250)
        basic.clearScreen()
        basic.pause(250)
    }
    // Put actual compass cal routine here
    basic.showLeds(`
        . . # . .
        . . # . .
        . . # . .
        . . . . .
        . . . . .
        `)
    basic.pause(10000)
})
```

source/drivers/MicroBitCompassCalibrator.cpp
jaustin: Yep, that makes sense. @finneyj in the original implementation did you have any reason for the way it was originally? Or intention to modify these points?

source/drivers/MicroBitCompassCalibrator.cpp
jaustin: I think in the first instance allowing button A+B to skip the message is best, because we have a real issue communicating to people what's happening with the compass calibration. @jamesadevine suggested this yesterday too. Just having a single button makes chances of users 'missing' the message too high. All of us approach this knowing exactly what's going on/about to happen but lots of people will be seeing it for the first time. 

We did toy with "Tilt to fill" which I think is probably fine too. Anyone object?

jamesadevine: @jaustin I'm not sure the text demystifies the process anymore than an animation, it will still be confusing for first time users; I like the idea of an animation more.

jamesadevine: @jaustin @DavidWhaleIET Got any new users to test an animation vs. text?

jaustin:
We did fairly good testing on old text versus new text as part of the
ticket and the new process was much clearer (it tells you clearly what to
do) and text is what's documented in a lot of places including some
upcoming books.

I think skipping the instructions at all seems pretty harsh - quick
animation plus some text seems maybe better, but a spinning icon doesn't
say "fill up the screen" at all.

jamesadevine:
@jaustin Okie dokie, can't deny user testing! :smile:

source/drivers/MicroBitCompassCalibrator.cpp
jaustin: @jamesadevine  So, in the testing we did with kids of text only "Tilt to fill screen" didn't need explanation in the same way "Draw a circle" did because they *way* do draw a circle wasn't clear, but the 'tilt' part explains how to fill the screen here - so that means the text alone worked, without video or other docs. 

The translation issue of the text is nasty though :/ It's a good reason to prefer icons over text long run. As far as I know we don't have any localisable text inside the pre-built components of micro:bit at the moment, and I think that's a 'good thing' in many ways. Does anyone know of previous discussions around localisation of the base components (IE MicroPython and DAL)?

jaustin:
Right, after discussion with @finneyj here's a couple of changes
* You can now use the calibration algorithm by providing a set of data points to the MicroBitCompassCalibrator::calibrate() function
* I've taken Damien's suggestion on static const for the point list.

However, I haven't made it possible to skip the message: The issue is that if something *can* be skipped it likely will (the 'click any button to make the dialogue go away" effect) and as this message is necessary for someone to proceed it's important to keep it. The message doesn't take very long compared to the calibration step itself, either - though I know it can be trying. 

I've also been reminded that the language "Tilt to fill screen" is now included in a book, too, so I'm inclined to keep that last word because it is nice and clear.

We did have some other possibilities for ways to skip this: could we notice that someone's started tilting already and go from there? Could we make it HOLD A+B for, say, 1 second?  Ultimately I don't want to hold off more testing of the *algorithm* itself right now though.

pelikhan:
Is the calibration persisted accross flash operations? If you work with the compass and you have to see that message 20x in hour, you'll likely start using another sensor.

jaustin:
Not yet, but it might go into #318 'utility mode' and that would solve it.
This commit doesn't introduce that problem though, and it solves the problem of common incorrect calibration. Thing is, you still can't skip the actual cal process even if you can skip the message.

microbit-mark:
Is there a hold up with this PR?

microbit-mark:
Just nudging this one again. What are the next steps?

jaustin:
Righty, the following commit: c41d955c1080aa17fb383c099d24a8ebae45ad38 modifies the game to allow background data collection and when you hit 15 points it stops the message.... That also made it easier to re-scroll the message after a timeout, so I've done that too.

We still have the last issues outstanding on the calibration but I think I'd consider merging the UX change now and the algorithm later so we get things in time for the v1.0 MicroPython release.

jaustin:
@finneyj thinks that there's a chance the 100 scale factor on units is causing an overflow in pathological cases and so I'll try a build without the scaling. If this doesn't resolve the occasional bad calibration then we'll split Ux from algorithm to get the Ux improvements in more expediently.

jaustin:
This has now been merged as part of the work to support different compasses and accelerometers and mostly went in via https://github.com/lancaster-university/microbit-dal/commit/bbf56210ed3d1dc10a48cc9ea2fdcb8c0ad5cd30

PR #324, opened 2017-11-15
flash_write - Default scratch_address & Erase scratch when overwriting flash

microbit-sam:
Comments before `flash_write` suggest that function calls leaving the scratch_addr parameter blank will use the default scratch page address. Currently it's NULL and means that the scratch page cannot be used.

This PR checks for a NULL value and replaces it with the default scratch address defined in MicroBitConfig.h

```/**
  * Writes the given number of bytes to the address in flash specified.
  * Neither address nor buffer need be word-aligned.
  * @param address location in flash to write to.
  * @param buffer location in memory to write from.
  * @length number of bytes to burn
  * @param scratch_addr if specified, scratch page to use. Use default
  *                     otherwise.
  * @return non-zero on sucess, zero on error.
  *
  * Example:
  * @code
  * MicroBitFlash flash();
  * uint32_t word = 0x01;
  * flash.flash_write((uint8_t*)0x38000, &word, sizeof(word))
  * @endcode
  */```


microbit-sam:
Also now erases the scratch page before it's used to prevent the data getting mangled

finneyj:
Hi @microbit-sam.

Thanks for your contribution - looks good! Sorry for delayed uptake of your PR, lots of other work on that's kept me away from micro:bit...

I think the only potential issue here would be the change in semantics for existing code. I've been through all the microbit-dal uses of this function (all of them are in MicroBitFileSystem). They would all follow your changed code path, but I think these would be unaffected because MicroBitFileSystem handles its scratch pages in a round robin fashion anyhow. For any external code we may not be aware of, we're just aligning with the documented API... so I think we're good in both cases.

Issue #323, opened 2017-11-15
Use custom version of atoi() to eliminate dependency on newlib's locale and save RAM


dpgeorge:
As discussed over at the MicroPython port to the micro:bit https://github.com/bbcmicrobit/micropython/issues/363, the latest version of newlib (2.5.0) introduced some new locale handling code and has a large (~350 bytes) `__global_locale` data structure that is not constant and lives in RAM.  On the micro:bit 350 bytes of RAM is a lot and it would be good to reclaim this.

Upstream at newlib they say that the only way around this is to add an option to newlib to completely disable the locale stuff, see https://sourceware.org/ml/newlib/2017/msg01041.html

Since that option might not be feasible for us (it requires a custom newlib build) I found another way: the only place in the DAL that uses the locale-dependent code (at least in the MicroPython port) is in `MicroBitImage::MicroBitImage(const char *s)` when it calls `atoi()`.  By simply replacing `atoi()` with a call to a custom version of this function (eg `my_atoi()`) that has a very simple implementation, one can reclaim 364 bytes of RAM and around 600 bytes of flash/ROM.

Is this a modification (changing atoi to my_atoi) that would be accepted in the DAL?

finneyj:
Hi @dpgeorge 

Yes, of course. No problem at all - a natural home would be here: 
https://github.com/lancaster-university/microbit-dal/blob/master/source/core/MicroBitCompat.cpp

Do you have a simple implementation to hand? If not, I have a few hundred minions learning C at the moment that could sharpen their teeth on this... They'll do anything for a mars bar.

dpgeorge:
Please see above PR for an idea to remove atoi().  Since the code calling atoi() was almost parsing the number already, I just adjusted it so it did construct the integer value.

finneyj:
yup - now merged, so closing this off. Thanks @dpgeorge.

PR #325, opened 2017-11-15
Rewrite MicroBitImage parsing code so it doesn't depend on atoi()

dpgeorge:
Using the function atoi() pulls in a dependency from newlib, which in turn
pulls in locale support, and locale support requires about 350 bytes of
RAM.  This patch removes such a dependency by rewriting the MicroBitImage
parsing code so that it directly converts the input string to an integer
instead of building a buffer and passing it to atoi().

This patch saves about 640 bytes of code space and 364 bytes of RAM.  It
also reduces the amount of stack space used by the MicroBitImage
constructor.

See #323 for further discussion.

### DIFF
--- a/source/types/MicroBitImage.cpp
+++ b/source/types/MicroBitImage.cpp
@@ -117,10 +117,8 @@ MicroBitImage::MicroBitImage(const char *s)
     int count = 0;
     int digit = 0;
 
-    char parseBuf[10];
-
     const char *parseReadPtr;
-    char *parseWritePtr;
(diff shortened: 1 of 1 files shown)


dpgeorge:
I tested this with the simple-animation code in microbit-samples.  It seems to work.

finneyj:
Thanks @dpgeorge - looks good, thanks for the contribution.

dpgeorge:
Thanks @finneyj!

PR #326, opened 2017-11-16
Remove "throw" exception specifications because they are deprecated.

dpgeorge:
The use of such specifications are deprecated since C++11, see for example https://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11.  Having these specifications here leads to lots of warnings on modern C++ compilers (eg arm-none-eabi-gcc v7.2.0).


finneyj:
Thanks @dpgeorge. Sorry for delayed uptake of your PR, lots of other work on that's kept me away from micro:bit.

Looks good - thanks for your contribution.

PR #329, opened 2017-12-11
Panic when codal heaps exist but are out of memory

mmoskal:
As discussed on Skype. We'll need a release to pull this in.

### DIFF
--- a/source/core/MicroBitHeapAllocator.cpp
+++ b/source/core/MicroBitHeapAllocator.cpp
@@ -344,6 +344,12 @@ void *microbit_malloc(size_t size)
         }
     }
 
+#if CONFIG_ENABLED(MICROBIT_PANIC_HEAP_FULL)
+    // Native malloc has issues, so skip the desperate fall back.
+    if (heap_count > 0)
+	    microbit_panic(MICROBIT_OOM);
+#endif
+
     // If we reach here, then either we have no memory available, or our heap spaces
     // haven't been initialised. Either way, we try the native allocator.
 


finneyj:
Hi @mmoskal 

Thanks for raising this PR. 

I'm looking at this in a slightly different way today. :)  In general, I agree with you that inducing a panic() when the microbit-dal heap is full is generally a good idea (as for kids' programs its better to fail fast and deterministicaly). However I worry that this patch might adversely affect BLE enabled programs, which will already be running really close to the wire on RAM...

I'm therefore wondering if we should only enable this patch for builds where BLE is disabled... Presumably we could have a slightly different CONFIG setup for MakeCode builds when BLE is enabled, right?

mmoskal:
So you think there are some allocations that will actually succeed using the fallback when BLE is enabled? Do you have some examples?

finneyj:
Yes, I think this is a possibility. The microbit heap only allocates a fraction of the available space, leaving some for libc. This is defined here: https://github.com/lancaster-university/microbit-dal/blob/master/inc/core/MicroBitConfig.h#L119

Default is 75%, so this would lock out some memory from application use.  I would think the most common scenario would be one of @bluetooth-mdw's BLE builds...

If we think this is a significant issue, i'm tempted to back port the codal heap allocator, which doesn't have this distinction (libc heap is not used there), then we can enable this panic() uniformly.

mmoskal:
I guess the problem is that malloc() doesn't check for stack overflow, right?

finneyj:
@mmoskal 

The changes to the heap allocator here will prevent this from happening, but without any possible waste of RAM:

https://github.com/lancaster-university/microbit-dal/commit/0c746140a3506c2a1991b4789cb0edbcb19cd97d#diff-06c1c6fa10310b8366d8b544af73c226R310

So I'd like to close this PR if that's OK by you?

PR #331, opened 2017-12-18
Appended asterisk to device name characteristic when in pairing mode

smartyw:
It's sometimes useful to be able to detect if the connected micro:bit is in pairing mode. By appending "*" to the device name characteristic in the Generic Access Service when in pairing mode, connected applications can identify this condition.


smartyw:
See https://github.com/lancaster-university/microbit-dal/issues/332

martinwork:
Does this change the "model_number_string" (2A24) from "BBC micro:bit" to "BBC micro:bit [xxxxx]*"?

smartyw:
No, only the Device Name characteristic (UUID 2A00) is affected by this change.

image

smartyw:
This will not work as well as I'd hoped. iOS doesn't allow access of any sort to the Generic Access Service like Android does so this "solution" would only help non-iOS devices. 

I suggest rejecting this PR!

Let's continue discussion here:

https://github.com/lancaster-university/microbit-dal/issues/332

The preference then is to do something similar with the advertised name but this will definitely impact some apps.

Let's continue discussing

PR #334, opened 2018-01-09
Reset the radio module after frequency change

spauka:
The nRF module requires that the radio be reset after the frequency is changed. Until this is done, we will not be able to receive packets.


finneyj:
Thanks @spauka 

Looks good to me. You even fixed a minor race condition for free too while you were there. :)

I've merged this into the dal-integration branch ready for the next release and beta testing, so closing off this issue. Thanks for your contribution!

Issue #348, opened 2018-03-19
delete[] operator not implemented in the heap allocator


thomasdenney:
Whilst `new[]` is implemented, `delete[]` is not implemented [here](https://github.com/lancaster-university/microbit-dal/blob/master/inc/core/MicroBitHeapAllocator.h#L155). Therefore if you attempt to free memory allocated with `new[]` with `delete[]` the system implementation of `free` is used rather than `microbit_free`. Using `delete` works as this is defined to use `microbit_free` but is not recommend C++ style.

pelikhan:
@finneyj i beleive the new heap allocator fixed this.

PR #349, opened 2018-03-20
Implement the delete[] operator

thomasdenney:
Exactly the same as `delete` (in the same way that `new` is the same as `new[]`) but ensures correct C++ semantics.

Resolves https://github.com/lancaster-university/microbit-dal/issues/348

### DIFF
--- a/inc/core/MicroBitHeapAllocator.h
+++ b/inc/core/MicroBitHeapAllocator.h
@@ -157,6 +157,13 @@ inline void operator delete(void *ptr)
     microbit_free(ptr);
 }
 
+/**
+  * Overrides the 'delete[]' operator globally, and redirects calls to the micro:bit heap allocator.
+  */
+inline void operator delete[](void *ptr)
+{
+    microbit_free(ptr);
+}
 
 // Macros to override overrides the 'malloc' and 'delete' functions globally, and redirects calls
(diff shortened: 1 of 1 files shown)


finneyj:
Hi @thomasdenney 

Thanks for your contribution here.

To address this issue and other related issues, we've made the microbit heap allocator more transparent in the beta version of microbit-dal 2.1.0, so that it no longer relies on function level overloads to intercept memory allocations. We now fully replace the libc heap allocator.

I'm 99.99% sure that this means that the issue behind this PR will be fixed by this, so if you're happy with this, I'd like to close off this PR?

Relevant patch is here, if you'd like to review it:
https://github.com/lancaster-university/microbit-dal/commit/0c746140a3506c2a1991b4789cb0edbcb19cd97d

p.s. Apologies for leaving this PR hanging so long.

thomasdenney:
Looks good to me. Is it still the case that you have to include the relevant header, or will this be the default allocator regardless?

finneyj:
That's correct, you no longer need to include the header. The MicroBitHeapAllocator will be used by default. 

You can disable this and return to using the libc allocator via a compile time CONFIG option, if you so desire, but would lose the ability to have multiple heap regions and safety in interrupt contexts.

PR #354, opened 2018-05-04
Added Partial Flashing Service (#9)

microbit-sam:
@microbit-sam to write PR info

@finneyj


martinwork:
Should there be a new config option to control creation of the PF service in MicroBitBLEManager::init?
If so perhaps this PR shouldn't increase the default MICROBIT_SD_GATT_TABLE_SIZE.

martinwork:
The PF service uses the MemoryMap object but doesn't seem to use its ability to write the map to flash. The MemoryMap class doesn't have any code to read the map from flash.

Should there be a config option to control the inclusion of the MemoryMap flashing code and whether it shifts DEFAULT_SCRATCH_PAGE to reserve a flash page?

The PF service constructor creates a MemoryMap instance, but this doesn't do anything.

martinwork:
Should this PR change the version in module.json?

martinwork:
What is the reason for the change in MicroBitFlash::erase_page, where it doesn't now retry sd_flash_page_erase? The top while loop is pointless in the new version.

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -78,6 +79,9 @@ DEALINGS IN THE SOFTWARE.
 #define MICROBIT_BLE_STATUS_STORE_SYSATTR       0x02
 #define MICROBIT_BLE_STATUS_DISCONNECT          0x04
 
+#define MICROBIT_BLE_MODE_PAIRING     0x00
+#define MICROBIT_BLE_MODE_APPLICATION 0x01
+
```
finneyj: @microbit-sam Are these actually used? tracing through, this feels like it might be dead code/causing incorrect state in MicroBitPartialFlashingService line 150?

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -146,7 +150,7 @@ class MicroBitBLEManager : MicroBitComponent
       * @param enableBonding If true, the security manager enabled bonding.
       *
       * @code
-      * bleManager.init(uBit.getName(), uBit.getSerial(), uBit.messageBus, true);
+      * bleManager.init(uBit.getName(), uBit.getSerial(), uBit.messageBus);
```
finneyj: Looks like we need to preserve the docs from the old version here?

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -220,7 +224,7 @@ class MicroBitBLEManager : MicroBitComponent
     void stopAdvertising();
 
     /**
-     * A member function used to defer writes to flash, in order to prevent a write collision with 
+     * A member function used to defer writes to flash, in order to prevent a write collision with
```
finneyj: cleanup whitespace only changes please, to ad clarity in the commit logs? :)

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -289,12 +306,21 @@ class MicroBitBLEManager : MicroBitComponent
...
+    void showManagementModeAnimation(MicroBitDisplay &display);
+
     #define MICROBIT_BLE_DISCONNECT_AFTER_PAIRING_DELAY  500
-    unsigned long pairing_completed_at_time;   
+    unsigned long pairing_completed_at_time;
 
     int pairingStatus;
     ManagedString passKey;
     ManagedString deviceName;
+    uint8_t bleMode = MICROBIT_BLE_MODE_APPLICATION;
```
finneyj: @microbit-sam as per above - feels like this needs to reflect other states too.

inc/bluetooth/MicroBitPartialFlashingService.h
```diff
@@ -0,0 +1,106 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 British Broadcasting Corporation.
+This software is provided by Lancaster University by arrangement with the BBC.
```
microbit-sam: Drop BBC (c) notice

inc/bluetooth/MicroBitPartialFlashingService.h
```diff
@@ -0,0 +1,106 @@
...
+    // M:B Bluetooth stack and MessageBus
+    BLEDevice           &ble;
+    EventModel          &messageBus;
+
+    /**
+      * Writing to flash inside MicroBitEvent rather than in the ISR
+      */
+    void partialFlashingEvent(MicroBitEvent e);
+
+    // The base address to write to. Bit masked:  (0xFFFF0000 & region.endAddress) >> 16
```
finneyj: @microbit-sam This feels like a risky assumption, if user code happens to cross a 64K Flash boundary (which it could). think we need to fix this one before merge please. :)

inc/bluetooth/MicroBitPartialFlashingService.h
```diff
@@ -0,0 +1,106 @@
...
+
+    // Ensure packets are in order
+    uint8_t packetCount = 0;
+    uint8_t blockPacketCount = 0;
+
+    // Keep track of blocks of data
+    uint32_t block[16];
+    uint8_t  blockNum = 0;
+    uint16_t offset   = 0;
+
```
finneyj: @microbit-sam @martinwork 

Just a sanity check here - what's our assumed semantics for what happens if a BLE disconnect/reconnect happens during an active partial flash session? Do we assume that previous state is maintained, or do we assume that all bets are off and that we are restarting a new session? I think either has merit, but probably very important that both sides agree. :)

Thoughts?

inc/core/MicroBitComponent.h
```diff
@@ -68,6 +68,9 @@ DEALINGS IN THE SOFTWARE.
 #define MICROBIT_ID_MULTIBUTTON_ATTACH  31
 #define MICROBIT_ID_SERIAL              32
 
+#define MICROBIT_ID_PFLASH_NOTIFICATION 33
+#define MICROBIT_ID_PAIRING_MODE        34
```
finneyj: @microbit-sam  Seems to be unused - dead code?

inc/core/MicroBitConfig.h
```diff
@@ -79,15 +79,21 @@ DEALINGS IN THE SOFTWARE.
...
-#define KEY_VALUE_STORE_PAGE	                (PAGE_SIZE * (NRF_FICR->CODESIZE - 17))	
+#define KEY_VALUE_STORE_PAGE	                (PAGE_SIZE * (NRF_FICR->CODESIZE - 17))
 #endif
 
-#ifndef BLE_BOND_DATA_PAGE 
+#ifndef BLE_BOND_DATA_PAGE
 #define BLE_BOND_DATA_PAGE                      (PAGE_SIZE * (NRF_FICR->CODESIZE - 18))
 #endif
 
+#ifndef MEMORY_MAP_PAGE
```
finneyj: @microbit-sam unused?

inc/core/MicroBitConfig.h
```diff
@@ -79,15 +79,21 @@ DEALINGS IN THE SOFTWARE.
...
+#define MEMORY_MAP_PAGE                      (PAGE_SIZE * (NRF_FICR->CODESIZE - 19))
+#endif
+
+// Scratch moved from page 19 to page 20
+// MicroBitFileSystem uses DEFAULT_SCRATCH_PAGE to mark end of FileSystem
 #ifndef DEFAULT_SCRATCH_PAGE
-#define DEFAULT_SCRATCH_PAGE	                (PAGE_SIZE * (NRF_FICR->CODESIZE - 19))
+#define DEFAULT_SCRATCH_PAGE	                (PAGE_SIZE * (NRF_FICR->CODESIZE - 20))
```
finneyj: Can we move this back now?

inc/core/MicroBitConfig.h
```diff
@@ -130,7 +136,7 @@ extern uint32_t __etext;
 // For standard S110 builds, this should be word aligned and in the range 0x300 - 0x700.
 // Any unused memory will be automatically reclaimed as HEAP memory if both MICROBIT_HEAP_REUSE_SD and MICROBIT_HEAP_ALLOCATOR are enabled.
 #ifndef MICROBIT_SD_GATT_TABLE_SIZE
-#define MICROBIT_SD_GATT_TABLE_SIZE             0x300
+#define MICROBIT_SD_GATT_TABLE_SIZE             0x400
```
finneyj: @microbit-sam I think this service should be included by default, but can we check if this additional 256 bytes of RAM allocation is really needed? RAM is very tight with BLE enabled, and I'd like to not take away more for applications than strictly necessary.

inc/drivers/MicroBitFlash.h
```diff
@@ -73,7 +73,7 @@ class MicroBitFlash
       * Erase an entire page.
       * @param page_address address of first word of page.
       */
-    void erase_page(uint32_t* page_address);
+    uint8_t erase_page(uint32_t* page_address);
```
finneyj: can we add an @return doxygen comment please?

inc/drivers/MicroBitMemoryMap.h
```diff
@@ -0,0 +1,123 @@
...
+
+#ifndef MICROBIT_MEMORY_MAP_H
+#define MICROBIT_MEMORY_MAP_H
+
+#include "mbed.h"
+#include "MicroBitConfig.h"
+#include "ManagedString.h"
+#include "ErrorNo.h"
+
+#define MICROBIT_MEMORY_MAP_MAGIC       0xCA8E
```
finneyj: needed?

inc/drivers/MicroBitMemoryMap.h
```diff
@@ -0,0 +1,123 @@
...
+            this->startAddress = startAddress;
+            this->endAddress = endAddress;
+            memmove( this->hash, &hash, 8 );
+        }
+
+        Region(){
+          this->regionId = 0x0;
+          this->startAddress = 0x0;
+          this->endAddress = 0x0;
+          memset( this->hash, 0xDD, 8 );
```
finneyj: Fine - but be consistent with whatever the NULL case is for above?

inc/drivers/MicroBitMemoryMap.h
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+    };
+
+    struct MemoryMapStore
+    {
+        Region memoryMap[3];
```
finneyj: use NUMBER_OF_REGIONS default here?

inc/drivers/MicroBitMemoryMap.h
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    uint8_t regionCount = 0;
+
+    /**
+    * Function to update the flash with the current MemoryMapStore
+    *
+    * @param memoryMapStore The memory map to write to flash
+    */
+    void updateFlash(MemoryMapStore *store);
```
finneyj: Do we need this functionality?

inc/drivers/MicroBitMemoryMap.h
```diff
@@ -0,0 +1,123 @@
...
+      * @param region The Region to update in the MemoryMap. The name is used as the selector.
+      *
+      * @return MICROBIT_OK success, MICROBIT_NO_DATA if the region is not found
+      */
+    int updateRegion(Region region);
+
+    /**
+     * Function to fetch hashes from PXT build
+     *
+     * @return int  Boolean result of the search. 1 = Hashes Found; 0 = No Hash Found
```
finneyj: no return type, so update comment?

module.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "microbit-dal",
-  "version": "2.0.0-rc9",
+  "version": "2.0.0-pf",
```
finneyj: Please remove, as we'll bump the version after merge...

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -641,12 +642,14 @@ void MicroBitBLEManager::pairingMode(MicroBitDisplay &display, MicroBitButton &a
     ManagedString namePostfix("]");
     ManagedString BLEName = namePrefix + deviceName + namePostfix;
 
-    ManagedString msg("PAIRING MODE!");
+    ManagedString msg("M M!");
```
finneyj: remove?

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -673,13 +676,17 @@ void MicroBitBLEManager::pairingMode(MicroBitDisplay &display, MicroBitButton &a
 
     // Stop any running animations on the display
     display.stopAnimation();
-    display.scroll(msg);
 
-    // Display our name, visualised as a histogram in the display to aid identification.
-    showNameHistogram(display);
+    // Replaced by animation TODO remove
+    //display.scroll(msg);
```
finneyj: dead code?

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -765,6 +772,50 @@ void MicroBitBLEManager::pairingMode(MicroBitDisplay &display, MicroBitButton &a
...
+    const int mgmt_animation_w = 20;
+    const int mgmt_animation_h = 5;
+    const uint8_t mgmt_animation[] =
+    {
+         1,1,1,1,1, 1,1,1,1,1, 1,1,0,1,1, 1,0,0,0,1,
+         1,1,1,1,1, 1,1,0,1,1, 1,0,0,0,1, 0,0,0,0,0,
+         1,1,0,1,1, 1,0,0,0,1, 0,0,0,0,0, 0,0,0,0,0,
+         1,1,1,1,1, 1,1,0,1,1, 1,0,0,0,1, 0,0,0,0,0,
+         1,1,1,1,1, 1,1,1,1,1, 1,1,0,1,1, 1,0,0,0,1
+    };
```
finneyj: Change to 255 to be consistent with below?

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -790,3 +841,24 @@ void MicroBitBLEManager::showNameHistogram(MicroBitDisplay &display)
             display.image.setPixelValue(MICROBIT_DFU_HISTOGRAM_WIDTH - i - 1, MICROBIT_DFU_HISTOGRAM_HEIGHT - j - 1, 255);
     }
 }
+
+/**
+ * Restarts in BLE Mode
+ *
+ */
+ void MicroBitBLEManager::restartInBLEMode(){
+   KeyValuePair* BLEMode = storage->get("BLEMode");
```
finneyj: Can we rename to "RebootMode" or similar, so this functionality is more general purpose?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+#include "MicroBitPartialFlashingService.h"
+
+// BLE PF Control Codes
+#define REGION_INFO 0x00
+#define FLASH_DATA  0x01
+#define END_OF_TRANSMISSION 0x02
+
+// BLE Utilities
+#define MICROBIT_STATUS 0xEE
+#define MICROBIT_RESET  0xFF
```
finneyj: Would these #defines be more useful in the header file matching this class?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+    // Get characteristic handle for future use
+    partialFlashCharacteristicHandle = partialFlashCharacteristic.getValueHandle();
+    ble.gattServer().onDataWritten(this, &MicroBitPartialFlashingService::onDataWritten);
+
+    // Set up listener for SD writing
+    messageBus.listen(MICROBIT_ID_PFLASH_NOTIFICATION, MICROBIT_EVT_ANY, this, &MicroBitPartialFlashingService::partialFlashingEvent);
+
+    // Set up Memory map
+    // This will create the Memory Map and store it in flash
+    MicroBitMemoryMap memoryMap;
```
finneyj: Do we need to create this here? look unused?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+          buffer[2] = (memoryMap.memoryMapStore.memoryMap[data[1]].startAddress & 0xFF000000) >> 24;
+          buffer[3] = (memoryMap.memoryMapStore.memoryMap[data[1]].startAddress & 0x00FF0000) >> 16;
+          buffer[4] = (memoryMap.memoryMapStore.memoryMap[data[1]].startAddress & 0x0000FF00) >>  8;
+          buffer[5] = (memoryMap.memoryMapStore.memoryMap[data[1]].startAddress & 0x000000FF);
+
+          // End Address
+          buffer[6] = (memoryMap.memoryMapStore.memoryMap[data[1]].endAddress & 0xFF000000) >> 24;
+          buffer[7] = (memoryMap.memoryMapStore.memoryMap[data[1]].endAddress & 0x00FF0000) >> 16;
+          buffer[8] = (memoryMap.memoryMapStore.memoryMap[data[1]].endAddress & 0x0000FF00) >>  8;
+          buffer[9] = (memoryMap.memoryMapStore.memoryMap[data[1]].endAddress & 0x000000FF);
```
finneyj: @microbit-sam  @martinwork  Is this marshalling the address into big endian format so that it's more useful on the Android/IOS side?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+          // Region Hash
+          for(int i = 0; i < 8; ++i)
+              buffer[10+i] = memoryMap.memoryMapStore.memoryMap[data[1]].hash[i];
```
finneyj: memcpy() might be simpler here?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+          // Send BLE Notification
+          ble.gattServer().notify(partialFlashCharacteristicHandle, (const uint8_t *)buffer, 18);
+
+          // Set offset for writing
+          baseAddress = (memoryMap.memoryMapStore.memoryMap[data[1]].startAddress & 0xFFFF0000) >> 16; // Offsets are 16 bit
```
finneyj: update as per earlier comment about safety on 64K boundaries?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+          // Reset packet count
+          packetCount = 0;
```
finneyj: @microbit-sam Looks like we also initialise blockPacketCount, blockNum and offset to zero statically when an instance of this class is created... do we need to reset any of this state here too?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+        }
+        case MICROBIT_RESET:
+        {
+          /*
+           * data[1] determines which mode to reset into: MICROBIT_BLE_MODE_PAIRING or MICROBIT_BLE_MODE_APPLICATION
+           */
+           switch(data[1]) {
+             case MICROBIT_BLE_MODE_PAIRING:
+             {
+               MicroBitEvent evt(MICROBIT_ID_PFLASH_NOTIFICATION, MICROBIT_RESET ,CREATE_AND_FIRE);
```
finneyj: again, no need for CREATE_AND_FIRE

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+/**
+  * @param data - A pointer to the data to process
+  *
+  */
+void MicroBitPartialFlashingService::flashData(uint8_t *data)
+{
+        // Receive 16 bytes per packet
+        // Buffer 8 packets - 32 uint32_t // 128 bytes per block
+        // When buffer is full trigger partialFlashingEvent
+        // When write is complete notify app and repeat
```
finneyj: Tighten up comment please - seems a slightly out of date?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+            uint8_t flashNotificationBuffer[] = {FLASH_DATA, 0xAA};
+            ble.gattServer().notify(partialFlashCharacteristicHandle, (const uint8_t *)flashNotificationBuffer, sizeof(flashNotificationBuffer));
+            packetCount = blockPacketCount + 3;
+            blockNum = 0;
+            return;
+        }
+
+        // Add to block
+        for(int x = 0; x < 4; x++)
+            block[(4*blockNum) + x] = data[(4*x) + 4] | data[(4*x) + 5] << 8 | data[(4*x) + 6] << 16 | data[(4*x) + 7] << 24;
```
finneyj: Can we memcpy() here?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+        switch(blockNum) {
+            // blockNum is 0, set up offset
+            case 0:
+                {
+                    offset = ((data[1] << 8) | data[2]);
+                    blockPacketCount = packetNum;
+                    blockNum++;
+                    break;
+                }
+            // blockNum is 7, block is full
```
finneyj: blockNum is 3 rather then 7, given the packet format?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        // Add to block
+        for(int x = 0; x < 4; x++)
+            block[(4*blockNum) + x] = data[(4*x) + 4] | data[(4*x) + 5] << 8 | data[(4*x) + 6] << 16 | data[(4*x) + 7] << 24;
+
+        // Actions
+        switch(blockNum) {
+            // blockNum is 0, set up offset
+            case 0:
```
finneyj: Can we update to the relevant #define command codes rather than literals please?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+      // Create a pointer to the data block
+      uint32_t *blockPointer;
+      blockPointer = block;
+
+      // Write to flash
+      flash.flash_burn(flashPointer, blockPointer, 16);
+
+      // Update flash control buffer to send next packet
+      uint8_t flashNotificationBuffer[] = {FLASH_DATA, 0xFF};
```
finneyj: Can we bring out these ACK codes along with the other protocol #define

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+      uint32_t *flashPointer   = (uint32_t *) ((baseAddress << 16) + offset + 0x40);
+
+      blockPointer = block;
+      flash.flash_burn(flashPointer, blockPointer, 16);
+
+      // Once the final packet has been written remove the BLE mode flag and reset
+      // the micro:bit
+        MicroBitStorage storage;
+        storage.remove("flashIncomplete");
+        delete &storage;
```
finneyj: I don't think we should be attempting to delete something stack allocated?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+}
+
+const uint8_t  MicroBitPartialFlashingServiceUUID[] = {
+    0xe9,0x7d,0xd9,0x1d,0x25,0x1d,0x47,0x0a,0xa0,0x62,0xfa,0x19,0x22,0xdf,0xa9,0xa8
+};
+
+const uint8_t  MicroBitPartialFlashingServiceCharacteristicUUID[] = {
+    0xe9,0x7d,0x3b,0x10,0x25,0x1d,0x47,0x0a,0xa0,0x62,0xfa,0x19,0x22,0xdf,0xa9,0xa8
+};
```
finneyj: @microbit-sam @martinwork Before we commit these into the annals of time, can I just check these UUIDs are consistent and unique with others we've used? @smartyw I seem to recall you kindly allocated the others for us?

source/drivers/MicroBitFlash.cpp
```diff
@@ -55,9 +55,16 @@ extern "C" void btle_set_user_evt_handler(void (*func)(uint32_t));
 static bool evt_handler_registered = false;
 static volatile bool flash_op_complete = false;
 
+/*
+ * static void SWI1_IRQHandler(uint32_t evt)
+{
+       flash_op_complete = true;
+}
+*/
+
```
finneyj: remove dead code please?

source/drivers/MicroBitFlash.cpp
```diff
@@ -101,15 +108,19 @@ int MicroBitFlash::need_erase(uint8_t* source, uint8_t* flash_addr, int len)
...
         flash_op_complete = false;
         while(1)
         {
-            if (sd_flash_page_erase(((uint32_t)pg_addr)/PAGE_SIZE) == NRF_SUCCESS)
+            uint8_t ret = sd_flash_page_erase(((uint32_t)pg_addr)/PAGE_SIZE);
+            if (ret == NRF_SUCCESS) {
                 break;
+            } else {
+                return ret;
+            }
```
finneyj: @microbit-sam  Looks like some code paths don't explicitly return a value here? Can we either update to be consistent or revert to earlier "void" function prototype if the return value is no longer needed?

source/drivers/MicroBitMemoryMap.cpp
```diff
@@ -0,0 +1,179 @@
...
+  *
+  * @return MICROBIT_OK on success
+  *
+  */
+int MicroBitMemoryMap::pushRegion(Region region)
+{
+
+    // Find next blank Region in map
+    if(regionCount == NUMBER_OF_REGIONS){
+        return MICROBIT_NO_DATA;
```
finneyj: return MICROBIT_NO_RESOURCES please?

source/drivers/MicroBitMemoryMap.cpp
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+        // Add data
+        memoryMapStore.memoryMap[regionCount].startAddress = region.startAddress;
+        memoryMapStore.memoryMap[regionCount].endAddress   = region.endAddress;
+        memcpy(&memoryMapStore.memoryMap[regionCount].hash, &region.hash, 8);
+        memoryMapStore.memoryMap[regionCount].regionId     = region.regionId;
```
finneyj: Can this be resolved down to:

"memoryMapStore.memoryMap[regionCount++] = region;

??

source/drivers/MicroBitMemoryMap.cpp
```diff
@@ -0,0 +1,179 @@
...
+                memoryMapStore.memoryMap[2].hash[2] = (*magicAddress & 0xFF0000)   >> 16;
+                memoryMapStore.memoryMap[2].hash[3] = (*magicAddress & 0xFF000000) >> 24;
+
+                magicAddress = (uint32_t *)(magicAddress + 0x1);
+
+                memoryMapStore.memoryMap[2].hash[4] = (*magicAddress & 0xFF);
+                memoryMapStore.memoryMap[2].hash[5] = (*magicAddress & 0xFF00)     >>  8;
+                memoryMapStore.memoryMap[2].hash[6] = (*magicAddress & 0xFF0000)   >> 16;
+                memoryMapStore.memoryMap[2].hash[7] = (*magicAddress & 0xFF000000) >> 24;
+
```
finneyj: Maybe use some memcpy() action here to reduce code size?

source/bluetooth/MicroBitPartialFlashingService.cpp
smartyw: UUIDs for custom services don't get registered anywhere so the safest approach is to allocate a fresh UUID using your favourite UUID generator (e.g. uuid on Linux). Consistency of number ranges doesn't matter. More important is to ensure the new service and its UUID is documented but I guess you'll get to that later.

finneyj: Thanks @smartyw. Did we not have some spatial locality for the other UUIDs we used though?

bluetooth-mdw: "spatial locality". I like that! Yes, we did but I think that's more likely because most were all allocated at more or less the same time. I believe I just let Bluetooth Developer Studio allocate them.... though who knows.... it's now a long time ago. You can either take the risk of colliding with someone else's UUID that they allocated after the micro:bit services were originally put together or avoid that risk altogether with a freshly baked one. They don't have any meaning other than as a unique ID and we do have some that definitely have no "spatial locality" already like the Nordic ones. I vote for a fresh one. If we're voting. Oh, let's have a referendum on the UUID! That's a tried and tested way of making decisions with no possibility of unintended consequences! ;-)

inc/bluetooth/MicroBitPartialFlashingService.h
martinwork: If the BLE connection breaks, I think it will be safest to restart the partial flash from the beginning, especially as it is quick. The app currently assumes that it can always start/restart the process by connecting and sending the region requests, without needing to reset micro:bit, so long as the status request says "pairing mode".

inc/core/MicroBitConfig.h
martinwork: Note also the RAM allocated in member variables that are only used in pairing mode.

source/bluetooth/MicroBitPartialFlashingService.cpp
martinwork: The iOS app creates a uint32_t from the bytes with multiplications and additions.

source/bluetooth/MicroBitPartialFlashingService.cpp
martinwork: The existing microbit ones, except the (Nordic?) UART are of the format E95Dxxxx-251D-470A-A062-FA1922DFA9A8. The PF ones are the same format except they start e97d.

microbit-sam:
note for myself:

- [ ] Replace source/asm/CortexContextSwitch.s
- [ ] Reduce GATT table size 
- [ ] Update docs

finneyj:
Hi @microbit-sam 

I'd like to merge this in soon - how are you getting on with those "notes to self" changes listed above?

jaustin:
Closing because this has been merged into microbit-dal via the dal-integration branch

pelikhan:
is there a spec for the service? (i guess the code is the spec)

microbit-sam:
There's some (slightly out of date) docs here:
https://github.com/microbit-sam/microbit-docs/blob/master/docs/ble/partial-flashing-service.md

We're currently working through the backlog of python/dal/support docs every morning so this will be up to date and on the lancaster uni docs soon

pelikhan:
Is the Android sample still up to date?

PR #361, opened 2018-07-12
Add SoftDevice values for S130v1

jaustin:
These values are determined from
http://infocenter.nordicsemi.com/pdf/S130_SDS_v1.0.pdf
pg 41

It appears using S130v2 would allow even lower memory consumption

This is for use in conjunction with the
bbc-microbot-classic-gcc-s130
target, which also needs a fix before it will work.


finneyj:
Thanks @jaustin 

PR #362, opened 2018-07-24
Fixed sample period lookup issue in MAG3110 and MMA8653 drivers.

Taylor-Woodcock:
After performing many tests on various micro:bit revs. it was apparent that the 1.3 board was consuming 1600uA under an idle test using a lite version of the dal-integration branch. 

Under further inspection, the issue was found with the sample period lookup not being multiplied by 1000 when writing the configuration to the device. Also, I have set the default magnetometer sample rate back to 10Hz to match the values found in the current master branch.

This fix reduces the power consumption by around 700uA, leaving the 1.3 boards running at 900uA under this same idle test.

### DIFF
--- a/source/drivers/MAG3110.cpp
+++ b/source/drivers/MAG3110.cpp
@@ -99,7 +99,7 @@ int MAG3110::configure()
 
 
     // Bring the device online, with the requested sample frequency.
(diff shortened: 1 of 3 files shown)


finneyj:
Good stuff - thanks @Taylor-Woodcock.

Issue #152, opened 2016-05-31
Powering up Serial source triggers hang


cefn:
I have been developing the prototype a bit further and have the following code, which hangs if I power the CO2 sensor on after the micro:bit. 

If I power on the sensor first, and THEN start the micro:bit then it often (but not always) can run the code without a hang.

Note there is no usage of the serial in this at all, and the sensor doesn't send data all the time, only in response to a serial request so I don't expect a huge flood of data, (although I don't know if there's any kind of startup serial data sent).

The circuit is the micro:bit, the sensor attached as...
* GND: Common Ground
* 14:  3.3V UART TX
* 15:  3.3V UART RX

...and a tact switch attached as...
* Leg1: 3V
* Leg2: 16

In normal operation it scrolls 0 when unpushed and 1 when pushed. However, when the sensor is brought to life while the micro:bit is running, then the display goes permanently dark. It reliably hangs when the sensor is powered up after micro:bit reset. Sometimes a reset of the Microbit while the sensor is still powered on restores function, sometimes it hangs again. 

The code below was compiled based on the microbit dal import on the MBed compiler from a few days ago, plus the single-line patch to remove serial free as mentioned by @jamesadevine 

If this considered the same bug as previously, then happy to collapse them.

```cpp
#include "MicroBit.h"

MicroBit uBit;

int main()
{
    uBit.init();
    uBit.serial.baud(9600);
    uBit.serial.redirect(MICROBIT_PIN_P14, MICROBIT_PIN_P15);

    while(1)
    {
        uBit.display.scroll(uBit.io.P16.getDigitalValue());                
    }
   
}
```

finneyj:
@cefn - do you think this is related to serial at all? What if you take out the two line related to serial initialization?

```
uBit.serial.baud(9600);
uBit.serial.redirect(MICROBIT_PIN_P14, MICROBIT_PIN_P15);
```

Behaviour changing with power-on ordering can often be a symptom of a hardware issue. i.e. unstable power supply etc.?

cefn:
Sorry not to have specifically noted this. I recall completing the sanity-check step you mention before posting. That was exactly what motivated me to post.

However, the sensor has now been shipped to the client, (along with instructions for a working reset sequence), so I don't have access to recreate the circuit and replicate those steps with certainty. To investigate further, it might be worth me waiting until I'm reunited with the sensor, (or its equivalent, the MH-Z14A which is being shipped). Then I can answer your question directly.

[...]

PR #363, opened 2018-07-25
Always read from the serial to clear the RX interrupt

philipphenkel:
The Micro:bit crashes if the RX buffer is not initialized when data is received. The cause is that the RX interrupt is not cleared in this case.

This problem always appears after a redirect without any read operation.

Fixes https://github.com/Microsoft/pxt-microbit/issues/917
Potential fix for #152

### DIFF
--- a/source/drivers/MicroBitSerial.cpp
+++ b/source/drivers/MicroBitSerial.cpp
@@ -88,11 +88,12 @@ MicroBitSerial::MicroBitSerial(PinName tx, PinName rx, uint8_t rxBufferSize, uin
   */
 void MicroBitSerial::dataReceived()
 {
+    //get the received character
+    //Note: always read from the serial to clear the RX interrupt
+    char c = getc();
+
     if(!(status & MICROBIT_SERIAL_RX_BUFF_INIT))
         return;
-
-    //get the received character
-    char c = getc();
 
     int delimeterOffset = 0;
     int delimLength = this->delimeters.length();


DavidWhaleMEF:
@finneyj @jaustin Can we get this fix into the next DAL release please? It affects all high level languages when they use the serial port, and causes a hard crash (display driver clearly stops updating the LED multiplex as the LEDs go bright)

philipphenkel:
@jaustin I worked on package for MakeCode (pxt-makerbit) when I found out that the Microbit crashes always whenever serial data was received. Serial MP3 devices send data for example. I created a MakeCode program that just sends commands - without any read. The result is a crash because the serial redirect attaches to the RX interrupt and calls the function that my PR fixes. My fix is supported by the mbed documentation of serial.

In addition, we should think about to not attach always to the RX interrupt in case of redirect.

ryanteck:
I believe this issue is causing development issues for a new add-on we are designing for the Microbit. 

Issue seems to be that when sending data with the add-on plugged in the microbit resets after a few packets are sent which is when the  module replies back with an OK command which we ignore. To debug each stage I added a function to display on the microbit's display a number to indicate which line it was at. At a stage where it would then move onto the next step the display crashes and leaves only a few of the pixels of the stage it crashes at.

The same UART commands work fine in the latest beta of Micropython.

finneyj:
Thanks  @philipphenkel @ryanteck 

Looks good to me.

I've merged this into the dal-integration branch for beta testing, which will go live with the next release, so closing off this issue.

@ryanteck - would be great if you could confirm that this fixes the issue you are seeing with your peripheral. Please do report back here of you do, and reopen this issue if you still see problems.

ryanteck:
Seems to have fixed it, sorry about the delay in getting back. I'll report if I find any more issues @finneyj

Issue #332, opened 2017-12-18
Determine over Bluetooth is micro:bit is in pairing mode


smartyw:
**Background**

The Bit:Bot product has a couple of line detection sensors which share pins with the two micro:bit buttons A and B. Consequently if you power it up while it is on a surface, it typically boots into pairing mode. You avoid this by lifting the Bit:Bot off the ground so that there's enough light falling onto the line sensors and then power on.

Inevitably however, users will boot into pairing mode accidentally and not realise their mistake. This leads to a poor user experience.

** Detecting Pairing Mode over Bluetooth **

There are two scenarios where it would be useful for an application to be able to determine whether or not a micro:bit was in pairing mode.

1) Before connecting

It would be useful to be able to determine from Bluetooth advertising packet content that a micro:bit is in pairing mode. This seems problematic. Adding a special character to the end of the device name in advertising packets would work, but seems to break the Samsung application (and perhaps others). Other advertising packet fields could perhaps be used (e.g. the inclusion or exclusion of the Appearance field) but not all client APIs give access to all advertising packet fields. 

If we decided to implement a solution to this sub-issue, I think we'd need to use the device name advertising packet field and affected applications would need to be changed.

2) After connecting

This is less problematic. One approach is to add a "*" to the end of the device name characteristic in the Generic Access Service. Then, after connecting, a client application can read that characteristic value, which is always guaranteed to be present, and determine whether or (perhaps) not, the micro:bit is in pairing mode.

I've submitted a PR which implements this:

https://github.com/lancaster-university/microbit-dal/pull/331 

And here it is in use in Bitty Controller:

![screenshot_20171218-075038](https://user-images.githubusercontent.com/1387793/34095197-681a7a5e-e3c8-11e7-98c1-cc7d8074c86d.png)

smartyw:
This will not work as well as I'd hoped. iOS doesn't allow access of any sort to the Generic Access Service like Android does so this "solution" would only help non-iOS devices. 

The preference then is to do something similar with the advertised name but this will definitely impact some apps.

Thoughts?

martinwork:
I was thinking in terms of what characteristics an app can see, but it seems iOS and macOS do reveal the GAP device name as explained here:
https://developer.apple.com/documentation/corebluetooth/cbperipheral/1519029-name?language=objc

smartyw:
[...]

PR #373, opened 2018-08-20
Determine micro:bit model version and add to Device Information Service

microbit-sam:
The micro:bit can detect what kind of micro:bit it is: 1.3x, 1.5x using by determining which accelerometer is in use.

This is then appended to "BBC micro:bit" in the Device Information Service.

### DIFF
--- a/inc/bluetooth/MicroBitBLEManager.h
+++ b/inc/bluetooth/MicroBitBLEManager.h
@@ -150,7 +150,7 @@ class MicroBitBLEManager : MicroBitComponent
       * bleManager.init(uBit.getName(), uBit.getSerial(), uBit.messageBus, true);
       * @endcode
       */
-    void init(ManagedString deviceName, ManagedString serialNumber, EventModel &messageBus, bool enableBonding);
+    void init(ManagedString deviceName, ManagedString serialNumber, EventModel &messageBus, bool enableBonding, ManagedString microbitModel = ManagedString("BBC micro:bit"));
 
     /**
      * Change the output power level of the transmitter to the given value.
@@ -323,6 +323,8 @@ class MicroBitBLEManager : MicroBitComponent
      * This variable will be set to MICROBIT_MODE_PAIRING if pairingMode() is executed.
      */
     uint8_t currentMode = MICROBIT_MODE_APPLICATION;
+
+    char* getMicrobitModel();
 
 };
 
--- a/inc/core/MicroBitDevice.h
+++ b/inc/core/MicroBitDevice.h
@@ -38,6 +38,10 @@ DEALINGS IN THE SOFTWARE.
 #define MICROBIT_NAME_CODE_LETTERS              5
 #define MICROBIT_PANIC_ERROR_CHARS              4
 
+#define MICROBIT_MODEL_UNKNOWN                  "BBC micro:bit"
+#define MICROBIT_MODEL_1_3_X                    "BBC micro:bit v1.3x"
+#define MICROBIT_MODEL_1_5_X                    "BBC micro:bit v1.5"
+
 #include "MicroBitConfig.h"
 #include "MicroBitMatrixMaps.h"
 
--- a/inc/drivers/FXOS8700.h
+++ b/inc/drivers/FXOS8700.h
@@ -232,6 +232,13 @@ class FXOS8700 : public MicroBitAccelerometer, public MicroBitCompass
     virtual void idleTick();
 
     /**
+    * Returns which accelerometer is detected
+    *
+    * @return MICROBIT_ACCELEROMETER_XXX
+    */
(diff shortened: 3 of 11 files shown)


microbit-sam:
Example program here:

[microbit-samples-combined.hex.zip](https://github.com/lancaster-university/microbit-dal/files/2301965/microbit-samples-combined.hex.zip)

bluetooth-mdw:
What about the future though? Will this essentially be binary versioning i.e. if it's 1.3 or 1.5 depending on the accelerometer? What if there are other editions which retain the current accelerometer?

microbit-sam:
I think if we have to update the board to revise another sensor or similar in the future, we should be able to identify it by the ICs on board / something similar

It should be fairly simple to modify the `getModel` function in `MicroBit.cpp` to query and identify a micro:bit's model depending on any future changes

https://github.com/lancaster-university/microbit/pull/12/files#diff-6047755bfccc64c91f0d390dd2f1061eR253

microbit-sam:
@martinwork 
Have you got any thoughts on this?

martinwork:
It would be more convenient if this info was in the advertisement, but the problems that might cause for existing software parsing the advertised name have already been discussed!

Thinking about the above and @bluetooth-mdw 's comment, it would be good to make the format explicit now and specify that other info could be added, so apps can allow for that when parsing. How much space is available?

Integer version numbers are slightly easier to deal with than float or text. 13 and 15? I don't know the significance of these actual version numbers. If we started with 0 and 1 now, it would fit it into a byte for 700+ years at the current rate!

bluetooth-mdw:
Agree wth @martinwork 

Also... the version information is of no use unless the distinction between micro:bit versions is going to be documented and there's a commitment to do that into the future. Where will this documentation live? "Documentation" may be the wrong word.... it's perhaps no more than a paragraph per version but whatever form it takes, it's needed.

martinwork:
Flags for: C++/makecode; flashIncomplete; compass calibrated?
4 binary flags could be encoded as a hex digit: F rather than 1111.

Or is this sort of state info not appropriate in the model characteristic?

martinwork:
getMicrobitModel is declared in MicroBitBLEManager.h but not defined in MicroBitBLEManager.cpp.

MICROBIT_MODEL_XXX string macros are defined in MicroBitDevice.h but only used in MicroBit::getModel() in lancaster-university/microbit.

MicroBit::getModel() returns a string. Would it make sense to have a function that returns a simple version number for use in code and turn this into a string for calling MicroBitBLEManager::init? Could the string conversion avoid including more text constant instances containing "BBC micro:bit"?

It looks like MICROBIT_BLE_MODEL is no longer used.

microbit-sam:
Does this need to be implemented in DAL 2.1.0 for anything in iOS app @martinwork ?

It sounds like this should be reworked slightly and we should all agree on a format before this goes forward, but just that this is a "nice to have" feature rather than a requirement when 1.5 is out

- 1.3x/1.5x are used rather than an integer to map directly to the board variant
- Mode flags (also discussed in #332) are an interesting idea but could a changing name confuse some users?
- getMicrobitModel(): Ok, need to remove that
- MICROBIT_MODEL_XXX string macros live in MicroBitDevice.h with all the other macros which makes sense to me
- MicroBit::getModel() returns a string: You're right, a version number is probably more useful and should be implemented

martinwork:
The only use the app has for this is that if it sees "1.3", it disables the alert for old hex files, which tells the user they might need an update to work with their micro:bit. Without the version, the app will always alert for old hex files. I still think integer version numbers would be better, but I have assumed "1.3" in the app now, so let me know if you change it! Most users will not see the model string. It doesn't need to be confusing, if it is specified and documented that it could change, but I don't know if it can or should be changed.

martinwork:
whatAmI (maybe should be whoAmI) is implemented using a virtual function.

Would it use less RAM and be simpler to add a single byte member variable in MicroBitAccelerometer  for the subclasses to store their ID, returned by a non-virtual function in MicroBitAccelerometer?

Maybe there are enough unused bits in MicroBitComponent::status?

inc/bluetooth/MicroBitBLEManager.h
```diff
@@ -150,7 +150,7 @@ class MicroBitBLEManager : MicroBitComponent
       * bleManager.init(uBit.getName(), uBit.getSerial(), uBit.messageBus, true);
       * @endcode
       */
-    void init(ManagedString deviceName, ManagedString serialNumber, EventModel &messageBus, bool enableBonding);
+    void init(ManagedString deviceName, ManagedString serialNumber, EventModel &messageBus, bool enableBonding, ManagedString microbitModel = ManagedString("BBC micro:bit"));
```
pelikhan: Any way to avoid using heap memory to store this info?

PR #378, opened 2018-08-26
Fix MicroBitPartialFlashingService blockPacketCount increment.

martinwork:
If the 5th packet goes missing, the recovery code would have set packetCount based on the wrong blockPacketCount, expecting the same packet numbers again.


finneyj:
Thanks @martinwork 

Issue #372, opened 2018-08-17
Tight loop checking gestures does not recognise shake


microbit-carlos:
We had reports of this code in MicroPython v1.0.0-rc.2 not registering the Shake gesture (while the old v0.9 doesn't have this issue):

```python
from microbit import *
while True:
    print(accelerometer.current_gesture())
    sleep(200)
```

The MicroPython method `accelerometer.current_gesture()` basically reads the `accelerometer.getGesture()` uBit method:
[@bbcmicrobit/micropython/source/microbit/microbitaccelerometer.cpp#L134-L139](https://github.com/bbcmicrobit/micropython/blob/da7ce603ff18763f42aacc0318d1118c1a6a65d7/source/microbit/microbitaccelerometer.cpp#L134-L139)

```c
mp_obj_t microbit_accelerometer_current_gesture(mp_obj_t self_in) {
    microbit_accelerometer_obj_t *self = (microbit_accelerometer_obj_t*)self_in;
    update(self);
    return MP_OBJ_NEW_QSTR(gesture_name_map[ubit_accelerometer->getGesture()]);
}
MP_DEFINE_CONST_FUN_OBJ_1(microbit_accelerometer_current_gesture_obj, microbit_accelerometer_current_gesture);
```

So I've tested the `accelerometer.current_gesture()` in a similar C++ program, and I can confirm that it has the same issue of not detecting the shake gesture, while it detects the UP/DOWN/LEFT/RIGHT correctly:

```cpp
#include "MicroBit.h"

MicroBit uBit;

const char* getAccGestureName(int eventValue) {
    switch (eventValue) {
        case  MICROBIT_ACCELEROMETER_EVT_NONE:       return ".";
        case  MICROBIT_ACCELEROMETER_EVT_TILT_UP:    return "\nUP";
        case  MICROBIT_ACCELEROMETER_EVT_TILT_DOWN:  return "\nDOWN";
        case  MICROBIT_ACCELEROMETER_EVT_TILT_LEFT:  return "\nLEFT";
        case  MICROBIT_ACCELEROMETER_EVT_TILT_RIGHT: return "\nRIGHT";
        case  MICROBIT_ACCELEROMETER_EVT_FACE_UP:    return "\nUP";
        case  MICROBIT_ACCELEROMETER_EVT_FACE_DOWN:  return "\nDOWN";
        case  MICROBIT_ACCELEROMETER_EVT_FREEFALL:   return "\nFALL";
[...]

PR #381, opened 2018-08-31
Register MicroBitAccelerometer::lastGesture on shake

microbit-carlos:
Info: https://github.com/lancaster-university/microbit-dal/issues/372#issuecomment-417731360

### DIFF
--- a/source/drivers/MicroBitAccelerometer.cpp
+++ b/source/drivers/MicroBitAccelerometer.cpp
@@ -311,6 +311,7 @@ void MicroBitAccelerometer::updateGesture()
 
     if (g == MICROBIT_ACCELEROMETER_EVT_SHAKE)
     {
+        lastGesture = MICROBIT_ACCELEROMETER_EVT_SHAKE;
         MicroBitEvent e(MICROBIT_ID_GESTURE, MICROBIT_ACCELEROMETER_EVT_SHAKE);
         return;
     }


source/drivers/MicroBitAccelerometer.cpp
```diff
@@ -312,6 +312,7 @@ void MicroBitAccelerometer::updateGesture()
     if (g == MICROBIT_ACCELEROMETER_EVT_SHAKE)
     {
         MicroBitEvent e(MICROBIT_ID_GESTURE, MICROBIT_ACCELEROMETER_EVT_SHAKE);
+        lastGesture = MICROBIT_ACCELEROMETER_EVT_SHAKE;
```
dpgeorge: I would set `lastGesture` before triggering the event, since the event handler may call `::getGesture()`.  This follows how it's done at the end of this function.

dpgeorge:
This looks good to me (except I'd set the variable before triggering the event).  I tested the change and it does fix the following test code:
```python
from microbit import *
while True:
    print(accelerometer.current_gesture())
    sleep(200)
```

source/drivers/MicroBitAccelerometer.cpp
microbit-carlos: Yes, good catch, I'll switch these two around.

finneyj:
Thanks @microbit-carlos @dpgeorge 

Looks good to me - perfect fix @microbit-carlos.

I've merged this into the dal-integration branch ready for the next release.

PR #389, opened 2018-09-24
Radio Patch: Correctly set TX Power, Update Frequency Limit

microbit-sam:


finneyj:
Thanks @microbit-sam. Looks great.

PR #391, opened 2018-10-02
Make microbit_heap_print() available externally

jaustin:
The heap printing in MicrobitHeapAllocator is really useful for
gauging how much memory is free at various points in the program.

Specifically, by calling this after the uBit object is created we
can start to measure a high-water-mark for RAM and track it over
releases, etc


finneyj:
yep.

PR #394, opened 2018-10-04
Allocated memory for the PFS buffer dynamically. Saves 60 bytes.

microbit-sam:
Please don't merge yet - still need to completely test Partial Flashing 

---

This PR only allocates the large buffer used in partial flashing when it's required


finneyj:
ACK - thanks.

pelikhan:
Probably best to have a C++ only test case too.

microbit-sam:
Sorry for the delay this week, been testing this afternoon and it's not resetting after flashing. Will update ASAP

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -200,6 +200,11 @@ void MicroBitPartialFlashingService::flashData(uint8_t *data)
 
         packetCount++;
 
+        // Reallocate block
+        if(block == NULL) {
+            (uint32_t*) malloc(16 * sizeof(uint32_t));
```
finneyj: Should this be ```block = ....``` ???

microbit-sam:
```suggestion
            block = (uint32_t*) malloc(16 * sizeof(uint32_t));
```

Issue #392, opened 2018-10-02
Reset in pairing mode not available for radio programs


pelikhan:
This event is ignored when BLE is turned off.
```
MicroBitEvent evt(MICROBIT_ID_PARTIAL_FLASHING, MICROBIT_RESET );
```

pelikhan:
@jaustin @finneyj @microbit-sam this feature only works with a bluetooth program., can we get it always on?

jaustin:
Looks like at the moment this is only getting handled in the PartialFlashingService but could potentially be registered in uBit https://github.com/lancaster-university/microbit/blob/master/source/MicroBit.cpp#L201 

@microbit-sam what if we added a handler in uBit only if we're not also bringing up BLE? Looks like it relies on a StorageObject in BleManager at the moment hence the dependency but we do also have that in uBit. It might be a bit of duplication but we could possibly factor it out to a static method that takes a stroageobject in BLEManager?

microbit-sam:
ack, will implement tomorrow

pelikhan:
PR in MakeCode: https://github.com/Microsoft/pxt-microbit/pull/1329

pelikhan:
@microbit-sam ping

microbit-sam:
Sorry for the delay @pelikhan

I've moved it into uBit, however the existing event listener is listening for when other listeners are created. We can moved the reset into pairing mode code into this listener, but it means that the event required will be:

`MicroBitEvent(MICROBIT_ID_MESSAGE_BUS_LISTENER, MICROBIT_ID_RESET_INTO_PAIRING);`

Which is probably a bad/confusing idea.. and it doesn't look possible to specify multiple event sources

@finneyj is this a bad idea?

finneyj:
@microbit-sam you should be able to use a full wildcard and register the uBit listener as (MICROBIT_ID_ANY, MICROBIT_EVT_ANY),  but then all the events in the micro:bit's universe would get piped through that function. ..

This is a processing / RAM usage tradeoff I guess... and the micro:bit is relatively processor rich.

microbit-sam:
Cool, once the PRs submitted against microbit and microbit-dal are reviewed and merged you will be able to trigger a reset into pairing mode with a `MicroBitEvent` (The event source currently doesn't matter, but may be useful to know the source in the future)

Example:
[...]

PR #395, opened 2018-10-05
Moved restart into pairing mode code

microbit-sam:
#392 

Moved the restart into pairing mode code to Microbit.cpp
It's #ifdef guarded to only compile if pairing mode exists in the build

I've left the source of the event here as MICROBIT_ID_PARTIAL_FLASHING. The event listener used now listens for any source but it may be useful in the future to know where the request to enter pairing mode has come from


microbit-sam:
Hey @finneyj - any update on this?

PR #401, opened 2018-10-24
Delay microbit_panic for 050 / 051 errors

microbit-sam:
This PR delays the microbit_panic call to the `requestUpdate()` function

A MicroBitAccelerometer / MicroBitCompass object is created rather than a device specific object, this allows the device to fall back to the default `requestUpdate()` function which now calls `microbit_panic()`

This function is only called if the user program requests data from the acc / mag, or sets up a listener for an event on the devices. If the program does not use the devices, the object is created but won't trigger the `requestUpdate()` / `microbit_panic()`

I've currently only tested this with a C++ build

------
@finneyj

### DIFF
--- a/source/drivers/MicroBitCompass.cpp
+++ b/source/drivers/MicroBitCompass.cpp
@@ -134,7 +134,8 @@ MicroBitCompass& MicroBitCompass::autoDetect(MicroBitI2C &i2c)
 
         else
         {
-            microbit_panic(MICROBIT_HARDWARE_UNAVAILABLE_MAG);
+            MicroBitCompass *unavailable = new MicroBitCompass(coordinateSpace, MICROBIT_ID_COMPASS);
+            MicroBitCompass::detectedCompass = unavailable;
         }
     }
 
@@ -357,6 +358,7 @@ int MicroBitCompass::getPeriod()
  */
 int MicroBitCompass::requestUpdate()
 {
+    microbit_panic(MICROBIT_HARDWARE_UNAVAILABLE_MAG);
     return MICROBIT_NOT_SUPPORTED;
 }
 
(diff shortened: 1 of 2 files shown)


jaustin:
@finneyj do you have any thoughts on this approach before we go much further in testing it? Would be good to get this into a MakeCode testing release now v1 is out.

source/drivers/MicroBitCompass.cpp
```diff
@@ -357,6 +358,7 @@ int MicroBitCompass::getPeriod()
  */
 int MicroBitCompass::requestUpdate()
 {
+    microbit_panic(MICROBIT_HARDWARE_UNAVAILABLE_MAG);
     return MICROBIT_NOT_SUPPORTED;
```
jaustin: @microbit-sam and @finneyj do you think we also need to panic in any other functions: is it misleading to allow things like clearCalibration() and configure() to succeed on the instance created when specific hardware isn't available? Same applies to the accelerometer too, I think. 

For real-world cases I think in all of these we're going to call requestUpdate() soon enough and get the actual failure, but are there cases where you can make 'illegitimate progress' in the absence device-specific instance and we should panic sooner?

microbit-sam: Thanks - yes please Joe!

finneyj:
done.

@microbit-sam - I'm about to dive into a tunnel of teaching until about 2:00...  Any chance you can do a few basic tests on master before i get back, then I can tag for you earlier this afternoon?

I've looked thorugh the commit log, and there's nothing much in there, but checking it compiles cleanly and your RADIO patches are good should do the trick, just to derisk things a bit?

microbit-sam:
Sure, no probs

microbit-sam:
Works for me! A tag for /beta would be great 😄

PR #410, opened 2018-12-05
Update CHANGELOG for v2.1.1

microbit-sam:
Version 2.1.1 (23/11/19 sam@microbit.org)
   - Partial flashing is no longer a default service in C++ builds
   - Make microbit_heap_print() available globally
   - Remove unused MicroBitPin objects
   - Add capability to set radio upper and lower frequency bands in yotta config
   - Fix radio power in C++ builds

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,13 @@ This file highlights all notable changes between tagged releases on the master b
 
 For a more detailed treatment of changes, please see the commit history of the github repository containing this file.
 
+Version 2.1.1 (23/11/19 sam@microbit.org)
+   - Partial flashing is no longer a default service in C++ builds
+   - Make microbit_heap_print() available globally
+   - Remove unused MicroBitPin objects
+   - Add capability to set radio upper and lower frequency bands in yotta config
+   - Fix radio power in C++ builds
+
 Version 2.1.0 (18/10/18 joe@comp.lancs.ac.uk)
    - Refactor of motion sensor devices to support a range of accelerometer/magnetometer 
      sensors through abstract high level classes and concrete device specific subclasses.


pelikhan:
So is the presence of the partial service an indicator that we are in pairing mode?

microbit-sam:
For MakeCode yes, but C++ builds don't include it in application or pairing mode

For a standard (I think) C++ `config.json` the micro:bit has DFU, Event, and Device Information Service in both application and pairing so I'm not sure it will be possible to determine mode by looking at services

martinwork:
Isn't the PF service present for both modes in a makecode hex with the BLE extension?

pelikhan:
my plan is to disable DFU/PF in application mode.

martinwork:
But it's currently present in both modes, isn't it? Do you mean to replace the application mode DFU/PF services with a slimmer capability to reboot to pairing mode or always require a manual A+B+Reset for flashing?

pelikhan:
Yes. We have to go in pairing mode for PF always so we might as well make it the same for DFU.

martinwork:
I'm not sure which of the 2 alternatives you are saying "yes" to. PF service has the facility to reboot to pairing mode, so the user doesn't always have to A+B+Reset to flash.

microbit-sam:
Currently yes, but the `ble-dal-integration` branch implements restarting to pairing mode as a MicroBitEvent. #409 
As the Event Service is available in application/pairing MakeCode builds this will hopefully replace the command found in the Partial Flashing Service. Does this sound right to you?

I'm heading up to Lancaster on Friday 14th to discuss Bluetooth with Joe, so hopefully we can merge into master before Christmas

Issue #340, opened 2018-02-14
Pairing mode starts advertising and shows the tick before pairing info is saved to flash


martinwork:
In MicroBitBLEManager, bleDisconnectionCallback calls advertise() immediately but defers saving of the pairing info to the idle callback. The delay is only a few milliseconds, but seems to be long enough for a waiting client device to reconnect before the save to flash. I noticed this seemed to cause a problem when pairing with Windows and I understand this may be causing a problem with some Android clients.

securitySetupCompletedCallback sets the flags that trigger the LEDs tick and the disconnect-and-save sequence in the idle callback. If "PAIRING MODE!" has finished scrolling, the delay between tick and save is probably long enough for a user waiting for the tick to reset micro:bit before the pairing info is saved.

The disconnect-and-save idle callback is not created until "PAIRING MODE!" has finished scrolling. If a user or client device does not need to wait for the LEDs histogram or pairing code display, the client device may be informed that pairing has completed 6 seconds or more before the pairing info is saved.

jaustin:
@microbit-sam can you take a look at this while you're looking at Pairing mode, please?

microbit-sam:
> In MicroBitBLEManager, bleDisconnectionCallback calls advertise() immediately but defers saving of the pairing info to the idle callback. The delay is only a few milliseconds, but seems to be long enough for a waiting client device to reconnect before the save to flash. I noticed this seemed to cause a problem when pairing with Windows and I understand this may be causing a problem with some Android clients.

Saving is deferred to avoid using the SD to write during a BLE event. As the data is eventually written to the flash using MicroBitFlash (asynchronously if SD is in use), saving could be invoked immediately and MicroBitFlash will continuously attempt to write until successful. I'm not sure if this will actually have much effect on the delay as the MB may end up waiting for a successful write, rather than waiting for an idle tick.
https://github.com/lancaster-university/microbit-dal/blob/2cff906f019944ae15d532157c1301c940f67862/source/drivers/MicroBitFlash.cpp#L154

> securitySetupCompletedCallback sets the flags that trigger the LEDs tick and the disconnect-and-save sequence in the idle callback. If "PAIRING MODE!" has finished scrolling, the delay between tick and save is probably long enough for a user waiting for the tick to reset micro:bit before the pairing info is saved.

[...]

PR #414, opened 2019-01-03
Pairing mode starts advertising and shows the tick before pairing info is saved to flash

microbit-sam:
#340

### DIFF
--- a/inc/bluetooth/MicroBitBLEManager.h
+++ b/inc/bluetooth/MicroBitBLEManager.h
@@ -95,7 +95,7 @@ struct BLESysAttributeStore
   * Class definition for the MicroBitBLEManager.
   *
   */
-class MicroBitBLEManager : MicroBitComponent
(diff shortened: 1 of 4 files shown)


microbit-sam:
@martinwork tried to tag you as a review but couldn't!

martinwork:
pairingComplete doesn't look right. I'm sorry - it looks like I have said something very similar "Looks good" before here: https://github.com/lancaster-university/microbit-dal/issues/340#issuecomment-441087587.

The tick or cross will show when MICROBIT_BLE_PAIR_COMPLETE is set. So the idea is to delay setting that for the success case until the info is saved in idleTick. 

If I am following right, this PR just moves it down a line. I think pairingComplete  needs to be like in this comment: https://github.com/lancaster-university/microbit-dal/issues/340#issuecomment-381731175

microbit-sam:
You're right, looks like I didn't actually finish this one off. Thanks Martin

PR #419, opened 2019-01-23
Reverse ordering of condition check [MicroBitSerial.cpp]

jamesadevine:
A byte would be read and then dropped if bufferIndex >= bufferLen. Reversing the condition check fixes this bug.


pelikhan:
Can we get this in?

PR #421, opened 2019-01-31
filesystem fix, radio packet size config

xenomote:
- Added a size getter for files
- Fixed several incorrect sanity checks
- Set the radio packet payload size to 254 bytes, as documented [here](http://infocenter.nordicsemi.com/pdf/nRF51_RM_v3.0.1.pdf), section 17.1.3
- Added or fixed overloaded operators to ManagedString and PacketBuffers


martinwork:
This would have been better as several separate PRs.

MICROBIT_RADIO_MAX_PACKET_SIZE should not be changed. It's not the radio maximum packet size. It's the maximum packet size for clients of the DAL. It's the size of the payload array in the FrameBuffer structure. Making it bigger will eat a chunk of RAM. While looking at this, I noticed that the range check on len in MicroBitRadioDatagram::send looks wrong.

MICROBIT_RADIO_DEFAULT_FREQUENCY should not be defined in MicroBitRadio.h. This became a config here https://github.com/lancaster-university/microbit-dal/commit/7aedfab59ac74cf74d8ec906f3aab9f5bcb1e6af.

I don't think a const operator[] needs to return a reference, but I could be wrong!

If PacketBuffer::getBytes is to be const, the pointer it returns should also be const. Could it have been created non-const on purpose, instead of creating const and non-const versions?

In ManagedString::ManagedString, checking length against strlen seems reasonable.

xenomote:
I understand it should have been several PRs but this is part of university project work with the microbit so the changes were made as I went along. Hence the default frequency is an accident, I'll remove that.

From what I could understand of the Nordic documentation and the existing code, it seemed that the MicroBitRadioDatagram info sits in the PAYLOAD section of the radio packet described in the Nordic docs, with the other fields S0, LENGTH and S1 unused it seemed that PAYLOAD could occupy the full 254 bytes. I may be wrong, but regardless, the original 32 bytes for payload seems needlessly limiting.

I removed the strlen check because it seemed redundant when the user provided length is available. Of course this puts the onus on the programmer to provide the correct length, but this doesnt seem unreasonable. Alternatively the length parameter could be dropped and strlen used instead, but that would be less efficient. I may have the wrong intentions but I'd like to hear your thoughts.

the smattering of const's was born of me trying to subclass the packetbuffer for my own uses, and in certain cases the const and pass by reference was required to satisfy the compiler (I can't remember the specifics because it was a while ago) but they were all quite necessary. If you need more detail then I can spend some time retracing to work out the details.

I'm not intimately familiar with C++ so I may be making some naive mistakes, but most of these changes seem justified

martinwork:
Maybe MICROBIT_RADIO_MAX_PACKET_SIZE could be a config option. You may be right about ManagedString. I am not sure what was intended.

xenomote:
Oh okay, I thought you were a member of the project :'D

The payload size is fairly important, as I said the 32 byte limit is prohibitively small for what my project (and I imagine many others') needs to achieve. I suppose I should split the PR but we'll see what the reviewers say.

The strlen check wasnt causing a problem as such, and I suppose I might have had the wrong idea about the intent. I guess the manual length might be used for limiting the input to a fixed maximum length, however in that case it doesn't make sense to discard the string entirely if it is too short, so I'll change it again to reflect this.

Thank you for your suggestions regardless

martinwork:
In 6 months time, an entry in the commit log saying "bugfixes and tweaks to filesystem" will not be as helpful as a sequence of focused fixes for specific issues, especially if the changes under that heading do not all relate to the file system. Making a PR as simple as possible makes it easy for a reviewer to check and accept.

pelikhan:
Add the config code to be able to change the radio packet size via a yotta config . Do not change defaults as they might break existing scripts.

PR #424, opened 2019-02-22
port codal fiber changes from codal for PXT GC support

mmoskal:
CC @abchatra


abchatra:
@mmoskal Can you please merge this? Also if possible create the right tags in the repos so we can point to it.

finneyj:
@mmoskal @abchatra 

Would you mind holding off on merging until I've had chance to review the changes to the scheduler fully? Some changes in this PR are pretty deep, and I feel another pair of eyes on it can only help. I'll try to find time in the next 24hrs...

abchatra:
Sure @finneyj. Sorry, I thought review was completed. (Green tick only means review requested 😞 ).

source/core/MicroBitFiber.cpp
```diff
@@ -65,6 +67,44 @@ static EventModel *messageBus = NULL;
...
+    if (queue && queue->prev)
+        microbit_panic(MICROBIT_HEAP_ERROR);
+    while (queue) {
+        if (*dest)
+            *(*dest)++ = queue;
+        (*sum)++;
+        queue = queue->next;
+    }
+}
+
```
finneyj: Looks fine, but just checking that sum is intentionally  not initialized to zero  in this function?  (I'm guess this is intentional from the variable names).

Can we get a doc comment block on this function?

source/core/MicroBitFiber.cpp
```diff
@@ -65,6 +67,44 @@ static EventModel *messageBus = NULL;
...
+    get_fibers_from(&dest, &sum, runQueue);
+    get_fibers_from(&dest, &sum, sleepQueue);
+    get_fibers_from(&dest, &sum, waitQueue);
+    __enable_irq();
+
+    // idleFiber is used to start event handlers using invoke(),
+    // so it may in fact have the user_data set if in FOB context
+    if (dest)
+        *dest++ = idleFiber;
+    sum++;
```
finneyj: Should this increment be inside the conditional? Looks like maybe it should?

source/core/MicroBitFiber.cpp
```diff
@@ -65,6 +67,44 @@ static EventModel *messageBus = NULL;
...
+  */
+int list_fibers(Fiber **dest)
+{
+    int sum = 0;
+
+    // interrupts might move fibers between queues, but should not create new ones
+    __disable_irq();
+    get_fibers_from(&dest, &sum, runQueue);
+    get_fibers_from(&dest, &sum, sleepQueue);
+    get_fibers_from(&dest, &sum, waitQueue);
```
finneyj: Looks ok, but don't forget we introduced a mutex semaphore as part of js-event-semantics... Each mutex has its own wait queue associated with it, so you could perhaps miss extracting those fiber context with this function? Would that cause a problem for your application of this?

https://github.com/lancaster-university/microbit-dal/blob/js-event-semantics/source/core/MicroBitFiber.cpp#L944

source/core/MicroBitFiber.cpp
```diff
@@ -314,6 +358,29 @@ void scheduler_event(MicroBitEvent evt)
...
+        if (forkedFiber != NULL) {
+#if CONFIG_ENABLED(MICROBIT_FIBER_USER_DATA)
+            forkedFiber->user_data = f->user_data;
+            f->user_data = NULL;
+#endif
+            f = forkedFiber;
+        }
+    }
+    return f;
+}
```
finneyj: Great - I've been wanting to factor this out into a function for a while... Thanks @mmoskal!

source/core/MicroBitFiber.cpp
```diff
@@ -412,28 +465,17 @@ int fiber_wait_for_event(uint16_t id, uint16_t value)
...
-            dequeue_fiber(f);
-            queue_fiber(f, &runQueue);
-            schedule();
-        }
+    // in case we created a new fiber, make sure to initialize its context
+    // in case schedule() isn't called immedietly afterwards
+    if (f != currentFiber) {
+        dequeue_fiber(f);
+        queue_fiber(f, &runQueue);
+        schedule();
```
finneyj: Hmmm.... I'm not sure this is entirely safe for this particular funciton. fiber_wake_on_event() is used to place the current fiber on a wait queue without entering the scheduler (such that race conditions can be avoided). i.e. 

fiber_wake_on_event(..)
// do something that will cause that event to happen at some later point
schedule();

or am I missing something?

source/core/MicroBitFiber.cpp
```diff
@@ -475,7 +523,7 @@ int invoke(void (*entry_fn)(void))
     if (!fiber_scheduler_running())
 		return MICROBIT_NOT_SUPPORTED;
 
-    if (currentFiber->flags & MICROBIT_FIBER_FLAG_FOB)
+    if (currentFiber->flags & (MICROBIT_FIBER_FLAG_FOB | MICROBIT_FIBER_FLAG_PARENT | MICROBIT_FIBER_FLAG_CHILD) || HAS_THREAD_USER_DATA)
```
finneyj: @mmoskal Can you please describe the conditions when HAS_THREAD_USER_DATA is set? Will this affect all MakeCode application fibers / event handlers? Wouldn't this affect RAM efficiency and context switch time somewhat?

I also wonder if a global CONFIG option to disable FOB might be simpler if that's the case? Just thinking of future maintenance here.  Would be good to get your thoughts.

source/core/MicroBitFiber.cpp
```diff
@@ -710,6 +764,20 @@ void release_fiber(void)
     // Remove ourselves form the runqueue.
     dequeue_fiber(currentFiber);
 
+    // limit the number of fibers in the pool
+    int numFree = 0;
+    for (Fiber *p = fiberPool; p; p = p->next) {
+        if (!p->next && numFree > 3) {
```
finneyj: OK, but can we bring this constant out as a CONFIG option please?

source/core/MicroBitFiber.cpp
```diff
@@ -754,6 +828,8 @@ void verify_stack_size(Fiber *f)
 
         // Recalculate where the top of the stack is and we're done.
         f->stack_top = f->stack_bottom + bufferSize;
+
+        currentFiber = prevCurrFiber;
     }
```
finneyj: is this just for the case where the code may be pre-empted by an interrupt that scans the fiber list? if so, it's a short piece of code, can we protect with a more traditional disable/enable IRQ?

source/core/MicroBitHeapAllocator.cpp
```diff
@@ -355,13 +366,17 @@ void* calloc (size_t num, size_t size)
...
     void *mem = malloc(num*size);
 
-    if (mem)
-        memclr(mem, num*size);
+    if (mem) {
+        // without this write, GCC will happily optimize malloc() above into calloc()
+        // and remove the memset
+        ((uint32_t*)mem)[0] = 1;
+        memset(mem, 0, num*size);
+    }
```
finneyj: cor - good spot.

finneyj:
Thanks for all the work on this @mmoskal.

Generally all looks positive. Happy to merge pending questions I've posted as line comments - so could you take a look at those?

source/core/MicroBitFiber.cpp
mmoskal: No, the idea is that you call this twice, once with dest==NULL to count, then you allocate, and then call it again to get the data.

finneyj: ah, ok.

source/core/MicroBitFiber.cpp
mmoskal: oh wait, the schedule() is there already on master; I don't quite understand why, but I think it's fine to align with codal-core

PR #428, opened 2019-03-10
MicroBitUARTService - wake up send() on disconnection

### LINKED ISSUE #407: MicroBitUARTService::send can get stuck when client disconnects
[uart-sample.zip](https://github.com/lancaster-university/microbit-dal/files/2592143/uart-sample.zip)

See attached sample. I have tested it using the Android "Serial Bluetooth Terminal" app. Sometimes when the client disconnects, the sendAccel fiber stops running (the numbers stop) though microbit hasn't crashed because the letters keep going. It may take a few attempts. It happens more easily if the showText fiber is paused by pressing A. It doesn't restart when the client reconnects.

I suspect send() gets stuck waiting for the the event from on_confirmation() which may not arrive if the client disconnects. This seems to be confirmed by adding serial trace before fiber_wake_on_event() and after schedule().

martinwork:
Using an onDisconnection callback, trigger the same event as on_confirmation to wake up send() if it's sleeping. Fixes #407.

### DIFF
--- a/source/bluetooth/MicroBitUARTService.cpp
+++ b/source/bluetooth/MicroBitUARTService.cpp
@@ -55,6 +55,14 @@ void on_confirmation(uint16_t handle)
 }
 
 /**
+ * Callback when a BLE GATT disconnect occurs.
+ */
+static void bleDisconnectionCallback(const Gap::DisconnectionCallbackParams_t *reason)
+{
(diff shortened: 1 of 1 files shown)


martinwork:
The event can be triggered by MicroBitBLEManager, saving 20+ bytes needed for the extra onDisconnection callback.

PR #429, opened 2019-03-10
MicroBitDFUService - use SERIAL_DEBUG->printf

### LINKED ISSUE #387: printf seems to malloc 428 + 1024 bytes and causes out-of-memory panic
https://github.com/lancaster-university/microbit-dal/blob/master/source/bluetooth/MicroBitDFUService.cpp#L105

printf("  ACTIVATING BOOTLOADER.\n");

This causes an out-of-memory panic if doing DFU with debug enabled.

I added a SERIAL_DEBUG->printf each side of it - only the first gets sent before the panic.

martinwork:
Avoids OOM panic. Fixes #387


pelikhan:
Debug only fix, low risk.

PR #430, opened 2019-03-10
MicroBitUARTService - wake up send() on disconnection - version 2!

### LINKED ISSUE #407: MicroBitUARTService::send can get stuck when client disconnects

martinwork:
From MicroBitBLEManager's bleDisconnectionCallback, trigger the same event as MicroBitUARTService's on_confirmation to wake up send() if it's sleeping. Fixes #407.

### DIFF
--- a/source/bluetooth/MicroBitBLEManager.cpp
+++ b/source/bluetooth/MicroBitBLEManager.cpp
@@ -29,6 +29,7 @@ DEALINGS IN THE SOFTWARE.
 #include "MicroBitStorage.h"
 #include "MicroBitFiber.h"
 #include "MicroBitSystemTimer.h"
+#include "NotifyEvents.h"
 
 /* The underlying Nordic libraries that support BLE do not compile cleanly with the stringent GCC settings we employ.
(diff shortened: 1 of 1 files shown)


jaustin:
@finneyj and @microbit-sam would be good to look at this small fix as this is an issue for the iOS app and the fix isn't invasive.

microbit-sam:
LGTM 👍

Issue #321, opened 2017-10-31
Measuring battery capacity


ferrygun:
Can you please guide how to measure the battery capacity like described in the FAQ page here: https://support.microbit.org/support/solutions/articles/19000069531-measuring-battery-capacity

Any sample code would be useful.

PR #449, opened 2019-07-03
Update to allow mbed and the mbed online compiler to build.

geowor01:
Update to allow mbed and the mbed online compiler to build.
This allows gcc_arm, armcc, and armclang to build the DAL.

Most changes are my personal changes in order to comply with the restrictions of the up to date compilers, and are meant to give equivalent behaviour to previously.


microbit-sam:
Linked to:
nrf51-sdk: https://github.com/lancaster-university/nrf51-sdk/pull/1
mbed-classic: https://github.com/lancaster-university/mbed-classic/pull/13
microbit-dal: https://github.com/lancaster-university/microbit-dal/pull/449

bluetooth-mdw:
Not looked at the detail but I see quite a lot of files have changed and I'm wondering what regression testing has been performed /  will be performed? I note the changes are "meant to give equivalent behaviour".... but would feel reassured if this had been proven. Thanks.

source/drivers/MicroBitCompassCalibrator.cpp
```diff
@@ -226,8 +226,8 @@ CompassCalibration MicroBitCompassCalibrator::spherify(Sample3D centre, Sample3D
 Sample3D MicroBitCompassCalibrator::approximateCentre(Sample3D *data, int samples)
 {
     Sample3D c,t;
-    Sample3D centre = { 0,0,0 };
-    Sample3D best = { 0,0,0 };
+    Sample3D centre;
+    Sample3D best;
```
microbit-carlos: The default constructor initialises the values to zero, but is there any specific reason to remove the `{0,0,0}`?
Also, to reduce the diff as much as possible, can we undo all the whitespace character changes?

source/core/MicroBitHeapAllocator.cpp
```diff
@@ -59,7 +59,13 @@ DEALINGS IN THE SOFTWARE.
 // A list of all active heap regions, and their dimensions in memory.
 HeapDefinition heap[MICROBIT_MAXIMUM_HEAPS] = { };
 uint8_t heap_count = 0;
-extern "C" int __end__;
+#if defined __CC_ARM || defined __ARMCC_VERSION
+#define HEAP_START Image$$RW_IRAM1$$ZI$$Limit
```
microbit-carlos: I have no experience with armcc, but this is a really weird-looking and cryptic symbol, is there perhaps a better one for this value?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -41,6 +41,11 @@ DEALINGS IN THE SOFTWARE.
 MicroBitPartialFlashingService::MicroBitPartialFlashingService(BLEDevice &_ble, EventModel &_messageBus) :
         ble(_ble), messageBus(_messageBus)
 {
+    packetCount = 0;
+    blockPacketCount = 0;
+    blockNum = 0;
+    offset   = 0;
```
microbit-carlos: Is this necessary for the compiler compatibility?

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -242,6 +243,7 @@ MicroBitBLEManager::MicroBitBLEManager(MicroBitStorage &_storage) : storage(&_st
  */
 MicroBitBLEManager::MicroBitBLEManager() : storage(NULL)
 {
+    currentMode = MICROBIT_MODE_APPLICATION;
```
microbit-carlos: If this is a bug fix, could you submit it as a different PR?
Same for `inc/bluetooth/MicroBitBLEManager.h`.

source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s
```diff
@@ -0,0 +1,292 @@
+@ The MIT License (MIT)
```
microbit-carlos: Is this file meant to replace https://github.com/lancaster-university/microbit-dal/blob/master/source/asm/CortexContextSwitch.s.gcc? If so should the old one be removed?

source/asm/TOOLCHAIN_ARM_STD/CortexContextSwitch.S
```diff
@@ -0,0 +1,293 @@
+; The MIT License (MIT)
```
microbit-carlos: Is this file meant to replace https://github.com/lancaster-university/microbit-dal/blob/master/source/asm/CortexContextSwitch.s.armcc ? If not, can `CortexContextSwitch.s.armcc` be moved to a subdirectory similar to this new file path pattern?

inc/drivers/MicroBitQuadratureDecoder.h
```diff
@@ -42,13 +42,13 @@ DEALINGS IN THE SOFTWARE.
...
+    int64_t         position;           // Absolute position
     MicroBitPin&    phaseA,             // Phase A input for decoding
                     phaseB;             // Phase B input for decoding
     MicroBitPin*    LED;                // LED output to assert while decoding
-    uint32_t        samplePeriod = 128; // Minimum sampling period allowed
-    uint16_t        faults = 0;         // Double-transition counter
-    uint8_t         LEDDelay = 0;       // power-up time for LED, in microseconds
+    uint32_t        samplePeriod;       // Minimum sampling period allowed
+    uint16_t        faults;             // Double-transition counter
+    uint8_t         LEDDelay;           // power-up time for LED, in microseconds
```
microbit-carlos: If these values are set in the constructor, do we need to set them here as well? If not I would rather reduce this PR as much as possible.

martinwork:
In places, this appears to be a development from an earlier version of the DAL. It looks like that's the only reason for the MicroBitBLEManager changes, for example.
Other files are purely white space changes.
Should we be adding .lib files? If they are necessary, maybe this should be in a branch or fork repo.
Will it still compile with yotta?
Why replace temporary arrays on the stack with new/delete heap allocations?
It's hard to make out what the essential changes are.

BLE_API.lib
```diff
@@ -0,0 +1 @@
+https://github.com/lancaster-university/BLE_API/#bbc2dc58f8da1d5a6d1882dddd5ce7399dc48889
```
microbit-carlos: I assume these lib files are now needed for mbed-cli, is there any way around them?
If not, we now have two sources of truth, the yotta json file and these individual lib files.
What can we do to keep these in sync?
Can these lib files use git tags instead of commit hashes?

jamesadevine:
I'm a bit nervous about most of these changes. I think these will require extensive regression tests. @jaustin

inc/drivers/MicroBitQuadratureDecoder.h
geowor01: I have set them in the constructor because I have removed them here due to the armc5 compiler error: Error: #321: data member initializer is not allowed.

source/asm/TOOLCHAIN_ARM_STD/CortexContextSwitch.S
geowor01: The makefile requires that file to exist there so that it can select which file to copy to `asm/CortexContextSwitch.s` at compile time.

source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s
geowor01: Same as previous.

source/bluetooth/MicroBitBLEManager.cpp
geowor01: It is not a bug fix, it is to compensate for the removal of the initializer due to `Error:  #321: data member initializer is not allowed` from the armc5 compiler.

source/bluetooth/MicroBitPartialFlashingService.cpp
geowor01: Same as others.

source/core/MicroBitHeapAllocator.cpp
geowor01: I do not know, I extracted this symbol from https://github.com/lancaster-university/mbed-classic/blob/master/targets/cmsis/TARGET_Freescale/TARGET_K20XX/TARGET_K20D50M/TOOLCHAIN_ARM_STD/sys.cpp.

source/drivers/MicroBitCompassCalibrator.cpp
geowor01: Same as the other initializer removals. I will remove the whitespace.

source/drivers/MicroBitQuadratureDecoder.cpp
```diff
@@ -205,7 +212,7 @@ void MicroBitQuadratureDecoder::poll()
 {
     NRF_QDEC->TASKS_READCLRACC = 1;
     position += (int32_t)NRF_QDEC->ACCREAD;
-    faults = min(UINT16_MAX, faults + NRF_QDEC->ACCDBLREAD);
+    faults = min(0xFFFF, faults + NRF_QDEC->ACCDBLREAD);
```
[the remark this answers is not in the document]
geowor01: It doesn't appear to be for the armc5 compiler no. Where would be best to define it?

source/types/MicroBitImage.cpp
```diff
@@ -854,7 +858,7 @@ MicroBitImage MicroBitImage::crop(int startx, int starty, int cropWidth, int cro
         newHeight = getHeight();
 
     //allocate our storage.
-    uint8_t cropped[newWidth * newHeight];
+    uint8_t *cropped = new uint8_t[newWidth * newHeight];
```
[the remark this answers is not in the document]
geowor01: The updated versions of all compilers will not allocate a dynamically sized array on the stack, `expression must have a constant value`. Though looking at this I have changed the type without changing how it is used, which can't be right.

source/core/MicroBitHeapAllocator.cpp
microbit-carlos: I can see this is also defined here: https://github.com/lancaster-university/mbed-classic/blob/master/targets/cmsis/TARGET_NORDIC/TARGET_MCU_NRF51822/TOOLCHAIN_ARM_STD/sys.cpp#L15

As weird looking at it seems, I guess it is what it is ¯\\\_(ツ)\_/¯

BLE_API.lib
geowor01: They do not need to be inside the microbit-dal repository at all, they can be in the microbit repository or pulled in manually.

source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s
microbit-carlos: So the makefile needs the `source/asm/CortexContextSwitch.s.gcc` as well?

Right now we have two GCC asm files:
- `source/asm/CortexContextSwitch.s.gcc`
- `source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s`

So that it can copy one of them to `source/asm/CortexContextSwitch.s`?
How does it decide which one of the two to copy when using GCC?

source/drivers/MicroBitQuadratureDecoder.cpp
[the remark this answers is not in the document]
microbit-carlos: Top of the file I guess, we can move it somewhere else if somebody else requests it.

BLE_API.lib
microbit-carlos: Right but they have to be "somewhere", no?
If pulled manually does that mean extra steps when setting up the workspace?
I wouldn't mind having a script that parses the dependencies from the yotta json file and executes the right mbed-cli commands when setting up the workspace, what do you think the best place for a script like that would be?

source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s
geowor01: When using the makefile (yotta?) it ignored the `TOOLCHAIN_*` directory and checks the compiler flags to copy the correct `.s.gcc/armcc` file to `.s`.

source/types/MicroBitImage.cpp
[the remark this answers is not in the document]
geowor01: These are the main things I would want to test thoroughly.

BLE_API.lib
geowor01: Somewhere yes.

source/types/MicroBitImage.cpp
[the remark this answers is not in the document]
microbit-carlos: Is that only on armcc5? Might be another C++11 feature.

source/asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s
microbit-carlos: Where is this makefile or what tool/steps generates it?

geowor01: https://github.com/lancaster-university/microbit-dal/blob/master/source/CMakeLists.txt

microbit-carlos: Thanks!
So it looks like we are manually maintaining that CMake file anyway, and in that case we can move the original GCC and ARMCC assembly files to the new location required by mbed-cli, and update these lines to find them:
https://github.com/lancaster-university/microbit-dal/blob/8c2403e9643f28d19ecd852ed9916fbc090496d2/source/CMakeLists.txt#L88-L94

geowor01:
The scope of this PR will be reduced to only include the changes required for ARMC6 compatibility.

martinwork:
Could the changes for each syntax compatibility issue be grouped, so we can record what syntax to avoid in the future? Maybe a separate PR for the changes needed for the build system?

source/CMakeLists.txt
```diff
@@ -85,14 +85,6 @@ if (YOTTA_CFG_MICROBIT_CONFIGFILE)
...
 endif ()
 
-if(CMAKE_COMPILER_IS_GNUCC)
-  file(REMOVE "asm/CortexContextSwitch.s")
-  configure_file("asm/CortexContextSwitch.s.gcc" "asm/CortexContextSwitch.s" COPYONLY)
-else()
-  file(REMOVE "asm/CortexContextSwitch.s")
-  configure_file("asm/CortexContextSwitch.s.armcc" "asm/CortexContextSwitch.s" COPYONLY)
-endif()
-
```
microbit-carlos: Doesn't this need to be updated to move the `asm/TOOLCHAIN_GCC_ARM/CortexContextSwitch.s` or `asm/TOOLCHAIN_ARM_STD/CortexContextSwitch.S` into `asm/CortexContextSwitch.s`?

source/drivers/MicroBitCompassCalibrator.cpp
microbit-carlos: If the constructor code has been reverted (was there constructor code to initialise these to 0?), shouldn't these 3 lines be reverted as well?

geowor01: There wasn't constructor code to initialise these as they are initialised to zero by default.

jamesadevine: @geowor01 I don't think there are any guarantees that struct fields are zero initialised unless they are placed in bss (globals).

geowor01: @jamesadevine The default constructor sets the value to zero. https://github.com/lancaster-university/microbit-dal/blob/8c2403e9643f28d19ecd852ed9916fbc090496d2/inc/types/CoordinateSystem.h#L81

jamesadevine: @geowor01 Ah cool :smile:, my bad!

source/CMakeLists.txt
geowor01: Yotta appears to pay attention to TOOLCHAIN_* directories, as there are plenty in mbed-classic which cause no conflicts, though I do need to modify the line below to point at the new location.

geowor01: So in summary, yes.

geowor01:
Commits https://github.com/lancaster-university/microbit-dal/pull/449/commits/5b015a324cdb70e6a548110bbd704fcb4de95b71 and https://github.com/lancaster-university/microbit-dal/pull/449/commits/4f1a31fa2236f322e31bf910fa1066509b126e39 are enough to allow mbed-cli to work with gcc-arm.

PR #452, opened 2019-09-30
Clarify what dal stands for

nealmcb:
Device Abstraction Layer, it seems....


jamesadevine:
Thanks @nealmcb :smile:

PR #475, opened 2020-12-01
fix time for js-event-semantics

martinwork:
@jaustin I hope this is what you need!


martinwork:
Add "tick64" to yotta config.json:
```
{
    "microbit-dal":{
        "tick64":1
    }
}
```

PR #488, opened 2023-03-08
Add GitHub Actions Bloaty workflow to check/profile build size.

microbit-carlos:
> **Warning**
>
> This PR needs a GitHub access token configured in this repo as a "secret", as Yotta will otherwise get rate-limited when trying to fetch the dependencies from GitHub.
>
> A read-only fine grained token is enough (so there is almost not risk if it gets accidentally leaked) and the secret name should be `YOTTA_GITHUB_AUTHTOKEN`.

Equivalent to these Bloaty PRs added to CODAL:
- https://github.com/lancaster-university/codal-microbit-v2/pull/262
- https://github.com/lancaster-university/codal-microbit-v2/pull/282

When this is merged each commit in this repo will build the microbit-samples BLE project with this commit and a parent commit (or branch base commit for PRs) and show the build size difference and also it'll show the size of largest components.

To view this info you need to go to the GitHub Action Job summary (the CI run page) for the commit, where all the info has been exported. For example, from my fork (as this doesn't run here yet):
https://github.com/microbit-carlos/microbit-dal/actions/runs/4363680069

For PRs a comment showing the size diff for the PR is also added automatically.

You can also trigger the workflow manually to see the diff size between arbitrary commits in this URL (once merged):
https://github.com/lancaster-university/microbit-dal/actions/workflows/size-diff.yml
image

### DIFF
--- a/.github/workflows/size-diff.yml
+++ b/.github/workflows/size-diff.yml
@@ -1,0 +1,178 @@
+name: Bloaty size diff
+
+on:
+  pull_request:
+    branches: '*'
+  push:
+    branches: '*'
+  workflow_dispatch:
+    inputs:
+      old-commit:
+        description: 'Base/old commit to compare against the branch HEAD or new-commit below'
+        required: true
+      new-commit:
+        description: '(Optional) New commit to use instead of branch HEAD'
+        required: false
+        default: ''
+
+jobs:
+  size-diff:
+    name: Run Bloaty
+    runs-on: ubuntu-latest
+    steps:
+      #########################
+      # Install the toolchain #
+      #########################
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
+      - name: Setup arm-none-eabi-gcc 10.3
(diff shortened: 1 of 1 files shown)


microbit-carlos:
One thing to keep in mind comparing random commits is that if an old commit from this repo (`lancaster-university/microbit-dal`) requires an old commit from `lancaster-university/microbit` or `microbit-samples`, then the build will fail, as the build script currently clones the latest `microbit-samples`, which will fetch whatever version of `microbit` it has configured in its `module.json`.


So, for example, in my fork I've manually triggered a run to compare with v2.1.0 vs v2.2.0-rc5, which fails because v2.1.0 requires an older version of the other repos:
https://github.com/microbit-carlos/microbit-dal/actions/runs/4363621337

But comparing v2.2.0-rc0 vs v2.2.0-rc6 works as they both compile with the latest `microbit-samples` and `microbit` repos:
https://github.com/microbit-carlos/microbit-dal/actions/runs/4363680069

image

